PR #1, opened 2016-05-05
Extracts content_loader from dmutils

Wynndow:
Copies the main content_loader file as well as associated tests.

Copies two methods from dmutils.config that content_loader depends on and adds them to converters.py. Updates the test file for them as they were previously tested implicitly.

Moves two methods from dmutils.formats as well as tests. Front end apps will need updating to import from here.


.cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
allait: Let's drop this file

dmcontent/converters.py
```diff
@@ -0,0 +1,46 @@
+from flask._compat import string_types
```
allait: We could change this to `six.string_types`, and then we don't need Flask as a dependency (need `six` though).

tests/conftest.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+from flask import Flask
+
+
+@pytest.fixture
+def app():
```
allait: We don't need this fixture anymore

requirements_for_test.txt
```diff
@@ -0,0 +1,8 @@
+-r requirements.txt
+Flask==0.10.1
```
allait: Don't need this anymore

scripts/run_tests.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
allait: Let's replace this script with `test` Makefile tasks, like we did in the apps

Wynndow: All done!

allait:
Looks great 👍

PR #2, opened 2016-05-09
Change permissions of push-tag.sh to be executable

Wynndow:
Jenkins was unable to run the script without it being executable.


allait:
👍

PR #3, opened 2016-05-09
Don't add number unit if question wasn't answered

allait:
Adding the unit to empty values makes the summary table consider number questions with unit to be always answered, so the unit is displayed in the summary instead of the 'Answer question' link.

Adding a check that the value is not empty before adding the unit.

#### Add number question unit to value for questions with assurance
Unit should be added to the value even if the question has assurance, since the template will render the value before assurance without adding the unit sign.

(Discovered with functional tests 🌤)


TheDoubleK:
This looks good :+1:

PR #4, opened 2016-05-11
Fix assertion packing/unpacking

TheDoubleK:
Changes made with @carolinegreen 

Fixes these two bugs:
 * https://www.pivotaltracker.com/story/show/118388103
 * https://www.pivotaltracker.com/story/show/119356913 

It seems like the way empty values are handled has changed and broken some things.  By not including the `assertion` key when there is no corresponding value things get better.


dmcontent/content_loader.py
```diff
@@ -396,8 +396,11 @@ def get_data(self, form_data):
 
     def _get_single_question_data(self, form_data):
         if self.id not in form_data and self.type != 'boolean_list':
-            if self.get('assuranceApproach') and '{}--assurance'.format(self.id) in form_data:
-                return {self.id: {'assurance': form_data.get('{}--assurance'.format(self.id))}}
+            if self.get('assuranceApproach'):
+                if '{}--assurance'.format(self.id) in form_data:
+                    return {self.id: {'assurance': form_data.get('{}--assurance'.format(self.id))}}
+                else:
+                    return {self.id: {'value': [] if self.type in ['list', 'checkboxes'] else None}}
```
allait: Do we need to add 'value' key here? Looking at the unpacking method, it should handle the missing key fine, and the API will return `answer_required` either way since `value` is a required key.

TheDoubleK: I'm not sure it matters either way? As you say, the behaviour is the same.

TheDoubleK: As an empty dictionary is returned from the final `else:` in this section I think it's nicer to leave it as returning an empty dictionary here too?

allait: We still need `{self.id: {}}`, so I don't think it's the same as the final else.

Looking into this a bit more, here's what I found:

* The schema `required` validator for missing keysalways returns `answer_required`.
* For `'value': None` getting `answer_required` depends on either "None" key removal (which looking at it I think we don't do for nested "value and assurance" objects) or type validator, which returns a different error depending on the expected type (`not_a_number` for numbers or `answer_required` for booleans, which is exactly why we need the empty list for checkboxes and radios – API won't rewrite `type` validator errors for JSON array fields, so it goes all the way to `minItems`, which also returns `answer_required`).

So `value: None` or `value: []` tries to use the `type` validator logic to get the expected `answer_required` error, which will need to change if we decide to have eg text questions with assurances. At the same time, not sending the key in the first place will return an error from `required` for any question type.

I think there are two ways to approach this: we can extend API `purge_nulls_from_data` to remove `None` keys from nested objects, streamlining the logic and allowing us to delete answers to the questions with assurances (in case any of them become optional). Or we can return an empty dictionary here, which won't allow removing an already existing answer, but also ends up with a more straightforward validation rules in the API and makes it clear that the existing value won't be removed (which is what `None` does for other fields).

allait:
👍

PR #5, opened 2016-05-27
Add before_summary_value support and split content questions

allait:
### Make ContentQuestion a factory for BaseContentQuestion objects
Makes it possible to split question class based on question types without changing the existing import paths or constructors in tests.

### Split existing question tests based on question type
Groups tests based on the example question type, so that it's easier to see which behaviours are tested for which question type.

### Prepend before_summary_value in ListSummary.value
Adds the values from question content `before_summary_value` to the question service data for display in summary tables.

Used by alphagov/digitalmarketplace-frameworks#283

### Extract Multiquestion logic into a separate class
Make Multiquestion a separate class and remove all logic related to `self.questions` from `Question`. This removes if statements from the `Quesiton` methods and makes sure that for each qustion with `type: multiquestion` we end up with a `Multiquestion` instance by adding the new class to the `QUESTION_TYPES` dictionary.

Most of the code is unchanged, the most tricky part of the refactoring is probably the multiple inheritance we need to use for the summary class. Since `MultiquestionSummary` needs both `QuestionSummary`
methods and `Multiquestion` methods instead of base `Question` ones, we need to create the diamond inheritance that Python resolves to `Multiquestion Summary` - `QuestionSummary` - `Multiquestion` -
`Question` method resolution order, which gives us exactly what we want.

### Keep the .form_fields order in .required_form_fields
Using set() doesn't guarantee a consistent order of form fields, which means we need to make sure tests are using `sorted()` when making assertions on the method return value. Keeping the order removes the need for this and since the number of fields per question is low there should be no performance difference.

### Add individual tests for each question types
Add a bit more interface tests for some question types. Mostly expands existing test classes for now, so this is more to set a structure for the question tests than a full suite.

### Split assurance `get_data` logic from question value `_get_data`
Makes `Question.get_data` method wrap `Question._get_data` to add assurance data to the value. For question types that could have assurance `._get_data` should be overridden in the subclass instead of `.get_data`.


dmcontent/content_loader.py
```diff
@@ -231,6 +223,7 @@ def get_field_names(self):
         ]
 
     def get_question_ids(self, type=None):
+        # FIXME should this ever return self.id?
```
TheDoubleK: Is this `FIXME` resolved?

allait: It's covered by a test of current behaviour, but it's not clear what the expectation is, so I've just marked this for the future.

TheDoubleK:
This makes me quite nervous, but looks legit!
:+1:

PR #6, opened 2016-07-21
[126772043] Filter descriptions by lot

idavidmcdonald:
### Reason and background information for this pull request
[Original story](https://www.pivotaltracker.com/story/show/122114497) was a small content change however the only way to do it was to introduce filtering descriptions by lot to the content loader (for which [this new story](https://www.pivotaltracker.com/story/show/126772043) was added).

First, a YAML schema was decided and added to the frameworks repository. [This PR](https://github.com/alphagov/digitalmarketplace-frameworks/pull/293) includes full details of the new schema. Examples of new schema:

```
description: Buyers will use the information suppliers provide here to create a shortlist.
```
or 
```
description:
    default: Buyers will use the information suppliers provide here to create a shortlist.
    digital-specialists: You can only offer 1 specialist for this opportunity. Buyers will use the information suppliers provide here to create a shortlist.
```

### In this pull request
- When provided with a `description` which is a dictionary of lot keys, we are able to filter this dictionary by lot. If a lot key is not found then we can fall back to the `default` key. If no `default` key exists in `description`, then we raise an error.
- Introduced description as a `@property` so we can raise an error if we try to access the description but it has not yet been filtered to it's chosen string.
- Existing behaviour of parsing a description which is a string or is `None` is not changed.
- Bumps version to 1.2.0
- Small refactor to move methods from `ContentManifest` to `ContentSection` class as they better belong there

### In particular, looking for any feedback on
- errors raised (if they could be subclassed etc)

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '1.1.0'
+__version__ = '1.2.0'
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -6,6 +6,7 @@ import re
 import os
 from collections import defaultdict, OrderedDict
 from functools import partial
+from six import string_types
 from werkzeug.datastructures import ImmutableMultiDict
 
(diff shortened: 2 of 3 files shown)


TheDoubleK:
Nice - all looks good :+1:

PR #7, opened 2016-10-24
Allow slug to be set manually in content

TheDoubleK:
Slugs are generally automatically generated, but in some cases (e.g. if there are extended unicode characters such as non-breaking hypens and spaces in the section name) we want to explicitly set a slug.

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '1.2.0'
+__version__ = '1.3.0'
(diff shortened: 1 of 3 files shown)


allait:
The change look good :+1:

One thing to note though is I've tried to break out some of the tests away from the massive "here's a manifest will all possible variations" ones, so it might be nice to have a separate `ContentSection` test for this.

TheDoubleK:
I like the idea of splitting out tests, but the slug-setting in the change here is specifically to do with the way that loading manifests works.  `ContentSection` must be created with a specific slug that is passed in to `__init__()` - it's the loader that sets the slug to be passed in, so needs to be part of the ContentLoader tests rather than ContentSection.

With that in mind, I might leave it how it is for now...

PR #9, opened 2016-11-07
Add support for nested lists/dicts in ContentMessage list fields

allait:
Content variation template uses a list of dictionaries for rendering framework agreement changes so we need to support arbitrary nesting of structures in ContentMessage.

The logic remains mostly the same:
* render simple fields right away
* wrap dictionaries with `ContentMessage`
* render/wrap all items inside a list individually


pcraig3:
👍

PR #10, opened 2016-11-09
Add markdown flag to TemplateField

allait:
Allows overriding single/multi line check to enabe/disable markdown processing for a given field.


pcraig3:
Let's do it. 👍  

```

PR #11, opened 2016-11-16
Add DynamicList question

pcraig3:
## Document changes to content loader

In order to do nice-to-have questions with optional evidence
questions, we've created a new DynamicList question type.

1. `DynamicList` questions

`DynamicList` mostly works like a multiquestion, with a few key
differences:

 - `.filter()` method merges its original templated questions with
   brief data to generate new questions.  This means we can
   leave a question's "question" field as "{{ item }}" and then
   pass a string into it from a brief just before rendering it.
 - `.filter()` method also rewrites question "id"s and "followup"s
   with trailing index numbers.
   eg `id: "yesno"` becomes `id: "yesno-0"` or whatever
 - `.get_data()` method transforms data that comes back from
   the form into something the API will understand, sanitizing
   values and grouping related questions together.
   eg `{"yesno-1": "true", "evidence-1": "yes"}` will become
   `{"dynamicListKey": [{}, {"yesno": True, "evidence": "yes}]}`
-  `.get_error_messages()` for a DynamicList looks for errors
   returned from the API that contain a "field".  The field is the
   name of the key that failed: it tells us whether the user didn't    answer the main question or whether they didn't answer the
   followup.  This means we can put the validation message in the
   right place.

2. tech debt type stuff

 - DynamicList questions only work with one followup.
   eg we can have `followup: "evidence"` but we can't have it
   require two or three other questions.
 - We haven't covered any of the stranger questions when making
   our DynamicList.  That is, DynamicLists can easily contain
   text, boolean, or large_textbox fields, but if you give them
   pricing or assurance questions, the whole thing will break.
 - `.get_question_as_section()` method was a quick hack that
   served us pretty well, but it's starting to break as we rely
   on it more and more.  Currently, it works fine for displaying
   data, but not for returning the processed form data to the
   rendered HTML form after a validation error.
 - `macros/toolkit_forms.html` multiquestion macro (in the
   supplier frontend) only supports nested questions where
   `type: text`
 - No way to unformat data 😫. After getting data back in the
   view (`request.form`), we call `.get_data()` on it to
   turn it into something the API will understand, but if
   we get a validation error, we aren't able to pass the
   data back to the form to repopulate our old values.
   It's a bit strange because the error messages show up
   where we expect them to, just not the old values.

**Also, there are no tests for our new stuff.**

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '2.1.0'
+__version__ = '2.2.0'
--- a/dmcontent/questions.py
+++ b/dmcontent/questions.py
@@ -240,8 +240,7 @@ class DynamicList(Multiquestion):
 
     def __init__(self, data, *args, **kwargs):
         super(DynamicList, self).__init__(data, *args, **kwargs)
-
-        self.type = 'multiquestion'
+        self.type = 'multiquestion' # same UI components as Multiquestion
 
     def filter(self, context):
         dynamic_list = super(Multiquestion, self).filter(context)
--- a/dmcontent/questions.py
+++ b/dmcontent/questions.py
@@ -287,9 +287,13 @@ class DynamicList(Multiquestion):
         q_data = {}
         for question in self.questions:
             q_data.update(question.get_data(form_data))
-        answers = sorted([(int(k.split('-')[1]), k.split('-')[0], v) for k, v in q_data.items()])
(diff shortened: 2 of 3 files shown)


galund:
Notes so far based on trying to write some unit tests:

 - `get_data` breaks in interesting ways if you don't call `filter` first. It might be nice if an unfiltered DynamicQuestion threw some decent exception to stop this.
 - even if you do filter, you get a `KeyError` if your context doesn't include the key specified in your question's`dynamic_field` (e.g. `brief.essentialRequirements` needs to be there). Assuming this should be an error at all, it'd again be nice to get a more clear exception about what you did wrong.

allait:
I've added tests to make the coverage happy. 👍

PR #13, opened 2016-12-23
Add `unformat_data` method to `Question` and `DynamicListQuestion`

idavidmcdonald:
This method unpacks service data to be used in a form. Dynamic lists data is stored differently to that passed to a form and therefore requires manipulation from one key into several top level keys.

So we can call `unformat_data` from any question without having to worry about if it is a dynamic list or not I have also added a method to the parent `Question` class that will simply return the data passed in.


allait:
Nice work 👍 I was worried this could get unwieldy, but it's a straightforward, clean approach.

PR #14, opened 2017-01-04
Return ordered dynamic list question errors

idavidmcdonald:
We want our question errors ordered by index. JSON schema errors on a list of objects should give us back our errors ordered by index. These errors are passed from the API to the content loader. Based on this assumption, we just need to maintain this order using an OrderedDict.

### DIFF
--- a/tests/test_questions.py
+++ b/tests/test_questions.py
@@ -1,4 +1,5 @@
 # coding=utf-8
+from collections import OrderedDict
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


tests/test_questions.py
```diff
@@ -349,27 +350,28 @@ def test_get_error_messages_unknown_key(self):
...
-            'example3-0': {
-                'input_name': 'example3-0',
-                'message': 'There was a problem with the answer to this question',
-                'question': u'First Need-3'
-            }
-        }
+        ]})
+
+        errors = list(ordered_dict.items())
+        assert errors[0] == ('example2-0', {
```
allait: You can still compare `errors` to a list of tuples with one assert here, which makes the check a bit cleaner:

```python
assert errors == [
    ('example2-0', ...),
    ...
]
```

allait:
Nice catch 👍

PR #17, opened 2017-01-13
Remove followup answers for False answer to dynamic list questions

idavidmcdonald:
For [this pivotal story](https://www.pivotaltracker.com/story/show/133440815)

Our form data may include followup answers which are non empty (if the user says yes to the question, adds followup answer and then changes their mind and says no to the original question). This would mean that `False` would be sent with the follow up answer in the form data. We wish to strip out the follow up answer in this case when using `get_data` as it is redundant information.

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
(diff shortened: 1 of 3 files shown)


Wynndow:
Not a review

Wynndow:
Comment

Wynndow:
👍

idavidmcdonald:
Three thumbs up? I must have really nailed this one!

PR #18, opened 2017-01-16
Add `get_previous_question_id` to `ContentSection` class

idavidmcdonald:
Essentially the opposite of `get_next_question_id` apart from the `question_id` argument is not optional as we assume no need to get the last question. Note, if this need does arise then this functionality can easily be added.


Wynndow:
👍 when tests go green

PR #20, opened 2017-02-08
[194] One question per page

samuelhwilliams:
* Adds functionality to generate slugs for all (non-multiquestion) questions
* Adds get_next/previous_question_slug, based off get_next/previous_question_id (which should probably be removed in the future).
* Version bump to 2.6.0.

### DIFF
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -188,12 +188,24 @@ class ContentSection(object):
         question = self.get_question_by_slug(question_slug)
         if not question:
             return None
+
+        if question.get('questions'):  # multiquestion
+            return ContentSection(
+                slug=question.slug,
+                name=question.label,
+                editable=self.edit_questions,
+                edit_questions=False,
+                questions=question.questions,
+                description=question.get('hint'),
+                _context=self._context
+            )
+
         return ContentSection(
             slug=question.slug,
             name=question.label,
-            editable=self.edit_questions,
-            edit_questions=False,
-            questions=question.questions,
+            editable=self.editable,
+            edit_questions=self.edit_questions,
+            questions=[question],
             description=question.get('hint'),
             _context=self._context
         )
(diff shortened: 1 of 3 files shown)


dmcontent/content_loader.py
```diff
@@ -188,12 +188,24 @@ def get_question_as_section(self, question_slug):
...
         if not question:
             return None
+
+        if question.get('questions'):  # multiquestion
+            return ContentSection(
+                slug=question.slug,
+                name=question.label,
+                editable=self.edit_questions,
+                edit_questions=False,
+                questions=question.questions,
```
allait: I think we can merge this with the following constructor into a single one using`questions=question.get('questions', [question])` here.

Editable flags should be the same for both. For G9 sections we need to set `editable=False` and `edit_questions=True`, meaning there should be no way to edit the section itself, only individual questions are editable. That's true for both multiquestions and normal questions and for both `edit_questions=False` since they're the leaf nodes (ie questions inside multiquestions aren't rendered as separate sections).

dmcontent/content_loader.py
```diff
@@ -249,6 +261,41 @@ def get_previous_question_id(self, question_id):
 
         return None
 
+    def get_next_question_slug(self, question_slug=None):
```
allait: Should we call these `get_next_question` and make them return the question object instead of the `.slug` attribute?

I also wonder if we can do anything to merge them with the `get_next_question_id` methods. It feels like the only difference between them is the attribute name we use for lookup.

dmcontent/content_loader.py
samuelhwilliams: OK

tests/test_content_loader.py
```diff
@@ -793,10 +793,106 @@ def test_get_previous_question_id_for_section_with_no_questions_returns_none(sel
         })
         assert section.get_previous_question_id(None) is None
 
-    def test_get_question_as_section(self):
+    def test_get_next_question_slug(self):
```
allait: There's a slightly newer set of tests in `tests/test_content_section.py`, I wonder if it'd be better to add these there instead.

The problem with this file is that it takes a very coarse-grained approach, testing the entire thing from creating objects from the original datastructure to the method itself, which means they tend to act as a binary test suite – if there are any changes to the format most of these tests will fail.

dmcontent/content_loader.py
samuelhwilliams: It is, and it would be good to do that work, but I think it's relied upon elsewhere in the code and I don't feel we have time to dig it out right now. I intended to add it to tech debt board but don't have write access.

PR #21, opened 2017-02-10
[#138702601]ContentSection: add prefill flag

risicle:
Story https://www.pivotaltracker.com/story/show/138702601

Really just followed the path of `editable`. Though noted that `editable` handling itself is a bit under-tested.

### DIFF
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -131,6 +131,7 @@ class ContentSection(object):
             return ContentSection(
                 slug=section['slug'],
                 name=section['name'],
+                prefill=section.get('prefill'),
                 editable=section.get('editable'),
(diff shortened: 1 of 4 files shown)


dmcontent/content_loader.py
```diff
@@ -131,6 +131,7 @@ def create(cls, section):
             return ContentSection(
                 slug=section['slug'],
                 name=section['name'],
+                prefill=section.get('prefill'),
```
allait: Should this default to `False`? It will probably get treated like that either way.

risicle: I thought about that but I thought a) I would follow what `editable` does and b) someone someday might want to be able to discriminate between "not set" and "explicitly False"...

PR #22, opened 2017-02-14
slugification: handle unicode whitespace properly

risicle:
This was tripping up on (letting through) nbsp's.


allait:
Where is `_make_slug` used with unicode input in submissions? I think the approach we agreed on with @samuelhwilliams was to generate slug from `question.id`, which should be ASCII only, and later we'll try to adopt the same thing in briefs.

risicle:
I'm getting nbsp's coming through in section ids in @cathrooney's `g-cloud-9-supplier-declaration` frameworks branch. Possibly that's a copy-paste error, but it seemed simpler to just make it robust to this kind of thing than to go round lecturing people every time this happened...?

PR #23, opened 2017-02-14
Add get_source attribute for the benefit of the CSV downloader.

galund:
- ticket https://www.pivotaltracker.com/n/projects/904880/stories/138703715
 - requirement is to get 'readable' text, i.e. not rendered into HTML
 - actually, the source can still contain HTML, as much HTML is valid
   markdown, so this is really just a quick-fix


galund:
Agreed, however, as the JINJA processing might produce HTML, the ideal case might actually be to do all the processing with both, and then convert back to markdown at the end! Which might or might not be really weird :) thanks

PR #25, opened 2017-02-15
Follow up questions

allait:
## Add support for the followup questions to the multiquestion

Supports new followup YAML format.

Makes a separate _drop_followups method for the Multiquestions and
adds a call to it from `.get_data`.

It extends the method to support multi-input followup questions (eg
it should correctly drop all fields for a pricing followup), but it
still assumes that the original question data is stored in question
id key (which is true for most questions apart from multiquestions
and pricing, so they cannot have followups at the moment).

### Fix dynamic list filtering for schema generator

Schema generation calls `.filter` to create separate manifests for
separate framework lots, so that it can generate a schema for each
lot.

With the addition of DynamicList questions, filter now requires
dynamic data to generate the actual list of questions specific to
the brief, but that's not something we want to happen when listing
the questions for the schema generation. Schemas should contain
the original generic nested questions, so we need a separate way to
filter the manifests without doing the dynamic question replacements.

In order to achieve this, I'm adding a `static` flag to `.filter`,
which when set to True will make it skip the "dynamic" parts of
the filtering logic.

### Fix unicode error when printing questions with non-ascii characters

`__repr__` should return a string object in Python 2, so we need to
encode the template source beforehand.

### Add drop_followups call to ContentSection.get_data

G9 submissions are transforming multiquestions into sections before
rendering the page or processing form data, so ContentSection.get_data
method is replacing Multiquestion.get_data, which means we need to
implement drop_followups logic in both places at the moment (in
multiquestions for briefs and in sections for submissions).

### Add a reversed followup dict property for form templates

Form templates need a reversed followup structure: a list of question
ids that follow a certain question value. Since this is harder to do
in Jinja we add a helper property to Question that returns a reversed
followup dictionary.

### Make drop_followups return None values for top-level data fields

For nested questions (eg questions insidea a dynamic list array) we remove the question field
completely, since the top-level data key will be replaced anyway.

For multiquestions that are serialized to separate top-level keys we set the follow-up value
to `None`, so that it's replaced if the question was previously answered with a follow-up.

### DIFF
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -100,14 +100,14 @@ class ContentManifest(object):
     def get_next_edit_questions_section_id(self, section_id=None):
         return self.get_next_section_id(section_id, only_edit_questions=True)
 
-    def filter(self, context, static=False):
+    def filter(self, context, dynamic=True):
         """Return a new :class:`ContentManifest` filtered by service data
 
         Only includes the questions that should be shown for the provided
         service data. This is calculated by resolving the dependencies
         described by the `depends` section."""
         sections = filter(None, [
-            section.filter(context, static)
+            section.filter(context, dynamic)
             for section in self.sections
         ])
 
@@ -395,12 +395,12 @@ class ContentSection(object):
         for question in self.questions:
             question.inject_brief_questions_into_boolean_list_question(brief)
 
-    def filter(self, context, static=False):
+    def filter(self, context, dynamic=True):
         section = self.copy()
         section._context = context
 
         filtered_questions = list(filter(None, [
-            question.filter(context, static)
+            question.filter(context, dynamic)
             for question in self.questions
         ]))
 
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -100,14 +100,14 @@ class ContentManifest(object):
     def get_next_edit_questions_section_id(self, section_id=None):
         return self.get_next_section_id(section_id, only_edit_questions=True)
 
-    def filter(self, context):
+    def filter(self, context, static=False):
         """Return a new :class:`ContentManifest` filtered by service data
 
         Only includes the questions that should be shown for the provided
         service data. This is calculated by resolving the dependencies
         described by the `depends` section."""
         sections = filter(None, [
-            section.filter(context)
+            section.filter(context, static)
             for section in self.sections
         ])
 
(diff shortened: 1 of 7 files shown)


tests/test_content_section.py
```diff
@@ -190,3 +192,23 @@ def test_filtering_a_section_with_a_description_template(self):
...
         assert section.filter({"lot": "digital-outcomes"}).description == "description for outcomes"
+
+    def test_question_followup_get_data(self):
+        section = ContentSection(
+            slug='section',
+            name=TemplateField('Section one'),
+            prefill=False,
+            editable=False,
+            edit_questions=False,
+            questions=[Question({'id': 'q1', 'followup': {'q2': [True]}, 'type': 'boolean'}),
```
idavidmcdonald: All our tests rely on the follow up value being `True`. May be worth considering having a test that uses `False`. Would make our tests a bit more rigorous, however I'm guessing it's unlikely that we will actually see that in the frameworks repo so fine either way.

Similarly, the YAML seems to allow us to have multiple values that would trigger the follow up question. I believe followups can only be attached to boolean questions so I'm assuming this is just forward planning and we are happy not testing it.

allait: Good call, I've fliped the values here. Followups can be attached to radios and checkboxes now too.

dmcontent/content_loader.py
```diff
@@ -94,14 +94,14 @@ def get_next_section_id(self, section_id=None, only_editable=False):
     def get_next_editable_section_id(self, section_id=None):
         return self.get_next_section_id(section_id, True)
 
-    def filter(self, context):
+    def filter(self, context, static=False):
```
allait: Yeah, good call. I went with static since the only time it's used in a condition is `if static`, but I get how dynamic could be more clear. Changed to `dynamic=False`

allait:
@idavidmcdonald 

I've added support for checkboxes questions with followups: https://github.com/alphagov/digitalmarketplace-content-loader/pull/25/commits/a5227aed9d78fb59e22461f0c537233648caf1ab

tests/test_questions.py
```diff
@@ -148,6 +148,25 @@ def test_get_question_ids_by_type(self):
...
+    def question(self, **kwargs):
+        data = {
+            "id": "example",
+            "type": "boolean"
+        }
+        data.update(kwargs)
+
+        return ContentQuestion(data)
+
+    def test_followup_values(self):
```
idavidmcdonald: Should we have a similar test for checkboxes?

allait: The `followup:` YAML format is the same for both, and that's the only thing this property is using, so it's the same for radios and checkboxes.

idavidmcdonald:
👍

PR #26, opened 2017-02-21
Checkbox tree

galund:
Hierarchy, like List, but hierarchical.

 - https://trello.com/c/y79rP1KA/197-suppliers-choose-a-category-backend
 - To be used for taxonomy-style questions, specifically for our category/sub-category question.

### DIFF
--- a/dmcontent/questions.py
+++ b/dmcontent/questions.py
@@ -557,15 +557,15 @@ class IndexedTree(object):
 
     def add_node(self, value, label, parent):
         node = TreeNode(value, label, parent)
-        self.index[value].append(node)
+        self.index[node.value].append(node)
         return node
 
 
 class TreeNode(object):
     def __init__(self, value=None, label=None, parent=None):
         self.children = list()
-        self.value = value
-        self.label = label
+        self.value = value or label
+        self.label = label or value
         self.parent = parent
 
 
@@ -730,8 +730,9 @@ class HierarchySummary(QuestionSummary, Hierarchy):
             """
             filtered_options = []
             for option in options:
-                is_missing = option['value'] in missing_values
-                if option['value'] in selection or is_missing:
+                value = option.get('value', option.get('label'))
+                is_missing = value in missing_values
+                if value in selection or is_missing:
                     option = option.copy()
                     filtered_options.append(option)
                     option['missing'] = is_missing
--- a/dmcontent/questions.py
+++ b/dmcontent/questions.py
@@ -498,14 +498,14 @@ class Hierarchy(List):
             return {self.id: None}
 
         value_list = form_data.getlist(self.id)
-        value_list.extend(self._get_missing_values(value_list))
+        value_list.extend(self.get_missing_values(value_list))
         value_list.sort()
         return {self.id: value_list or None}
 
     def summary(self, service_data):
-        return ListSummary(self, service_data)
+        return HierarchySummary(self, service_data)
 
-    def _get_missing_values(self, selected_values):
+    def get_missing_values(self, selected_values):
         """
         Recursively retrieves un-selected parent categories of the
         passed-in selection, as a set of 'value' strings.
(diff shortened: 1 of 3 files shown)


dmcontent/questions.py
```diff
@@ -606,7 +652,7 @@ def value(self):
 class ListSummary(QuestionSummary, List):
     @property
     def value(self):
-        # Look up display values for options that have different labels from values
+        # TODO look up display values for options that have different labels from values
```
TheDoubleK: `TODO`... but not right now?

dmcontent/questions.py
```diff
@@ -480,6 +479,53 @@ def summary(self, service_data):
...
+            for option in options:
+                value = option.get('value', option.get('label'))
+                if value in selected_values_set:
+                    expected_values.update(parents)
+                children = option.get('options')
+                if children:
+                    set_expected_values(children, parents + [value])
+
+        set_expected_values(self._data.get('options', []))
+        return expected_values - selected_values_set
```
TheDoubleK: Ah OK, it's used in the summary below.

Do we have an actual use-case for the `missing` flag there? Or it just feels like something that should be useful?

TheDoubleK:
We could probably squash some of these commits together before merging.

dmcontent/questions.py
galund: This was kinda a compromise position for @allait and me, the rationalisation being roughly:

 - really, these additional categories are a denormalization. Arguably, the frontend should just assume that they are selected - the place they actually _need_ to be selected is in the search engine
 - however, the indexer doesn't have access to the tree at the moment, so moving the denormalization there is rather hard
 - the next best place to put this might then be the API, because that would ensure a consistent view on the data. But that would require teaching the API about this as a special field, and no other fields get processing - either the data posted in is accepted, or not (i.e. we only have validation, nothing beyond that). And again, the API doesn't currently have access to the original framework data, so even if we wanted to have the additional categories added there, it's hard in the current structure.
 - that leaves us with having to do the denormalization in the frontends (i.e. in the content loader)
 - having the API validate that the denormalization _was_ applied means duplicating that content loader logic into the API, and again we don't actually have the data available that would allow us to do that anyway...
 - and therefore data _validly_ posted to the API may not have the denormalization applied (i.e. if someone posted to the API via some other method than the content loader).
 - and that's weird, but we can't do anything about it, and weird stuff shouldn't be silent, so here we are.

Ideally, I would find a way to write a pithy one-line comment somewhere that explained the above :-D

dmcontent/questions.py
galund: So the comment was copy-pasted way-back-when, but actually was wrong. So I'm correcting the comment (but in fact @samuelhwilliams has a PR for this on the way :) )

dmcontent/questions.py
allait: @galund I'd put it in the commit message related to this change.

This is a minor point, but I'd use an argument instead of relying on a closure. It makes clearer that this is a recursive function with an accumulator and will make it easier to move it out if we need to since it's self-contained.

So something like:
```python

def get_expected_values(expected_values, options, parents=None):
    ...
    if children:
        get_expected_values(expected_values, children, parents + [value])
    ...

return get_expected_values(set(), self._data.get('options', [])) - selected_values_set
```

dmcontent/questions.py
```diff
@@ -480,6 +479,51 @@ def summary(self, service_data):
...
+
+    def get_missing_values(self, selected_values_set):
+        """
+        Recursively retrieves un-selected parent categories of the
+        passed-in selection, as a set of 'value' strings.
+        :param selected_values_set: initially-selected categories (e.g. by the user)
+        :return: additional values that should also be selected
+        """
+
+        def update_expected_values(options, parents, expected_set):
```
galund: @allait is this ok now? - didn't want to make it return, as it felt confusing to both return _and_ update the passed-in object, and I don't want to make copies of the set everywhere (which I feel would be implied by having the return)

allait: Ah, I see your point. Since this is not a method, returning a value feels more natural, but you're right in that you'd have to copy to avoid mutation.

I think in this case it'd be ok to pass in `set()` as an argument directly and then there's no side-effect since we don't have a reference. You could also default `expected_values` to `None` and then replace `None` with `set()` in the function itself, avoiding the need to pass it in as the argument.

But that's probably only marginally better, so I'm happy with keeping this as is.

PR #28, opened 2017-02-28
Specify a default error message for 'answer required'.

galund:
The idea is that we should not have to repeat a standard error message in hundreds of question files, as we do at the moment.

Also added a full stop in the fallback error message, for consistency with the error messages we wrote into our YAML.

### DIFF
--- a/tests/test_questions.py
+++ b/tests/test_questions.py
@@ -455,30 +455,39 @@ class TestDynamicListQuestion(QuestionTest):
 
     def test_get_error_messages(self):
         question = self.question().filter(self.context())
-        ordered_dict = question.get_error_messages({'example': [
-            {'field': 'yesno', 'index': 0, 'error': 'answer_required'},
-            {'field': 'evidence', 'index': 0, 'error': 'answer_required'},
-            {'index': 1, 'error': 'answer_required'}
-        ]})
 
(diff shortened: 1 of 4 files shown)


tests/test_questions.py
```diff
@@ -441,30 +441,39 @@ def test_get_error_messages_unknown_key(self):
...
-                'input_name': 'evidence-0',
-                'message': 'There was a problem with the answer to this question',
-                'question': u'First Need-evidence'
-            }),
-            ('example', {
-                'input_name': 'example',
-                'message': 'There was a problem with the answer to this question',
-                'question': 'Dynamic list'
+
+        errors_and_messages = {
```
idavidmcdonald: Is there a particular reason you've tested this functionality on dynamic list questions?

galund: None afaik: this was the existing test, which I had to fix because I'd changed the message - so it made sense to then test both messages here, if I was changing it anyway. Don't know why that one was chosen.

samuelhwilliams: It would be nice to test this in a more general class as it might give the wrong impression if it lives here.

Don't think it matters too much either way, but probably isn't a lot of effort to move this test to somewhere more neutral.

Otherwise, 👍

PR #29, opened 2017-03-01
No longer persist parent values in the checkbox tree to the API.

galund:
- https://trello.com/c/y79rP1KA/197-suppliers-choose-a-category-backend
 - implements 'decision re validation / parent categories' (see ticket)


pcraig3:
Looks like a small enough change, and I'm also using this code already to locally in the supplier app and it's doing what I expect.

👍  once you do a version bump.

PR #30, opened 2017-03-03
Add support for subfield TemplateFields

samuelhwilliams:
* Lot descriptions are currently stored in services/lots.yml for each framework, roughly working like a radios question.
* Descriptions for each possible choice are stored in options[x].description.
* Adds a TEMPLATE_OPTIONS_SUBFIELDS variable to Question to specify which subfields under options should be turned into TemplateFields and rendered appropriately.
* This check in options subfields does not iterate down beyond the first level, so will not work with - e.g. - CheckboxTree.


TheDoubleK:
:+1: once up to date (will need a new version number).

PR #32, opened 2017-03-08
Drop everything except letters and numbers before generating slugs

samuelhwilliams:
* Drop everything except letters and numbers (regex \w) from ids/names before generating the slug.
* Bump version to 5.3.1

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '3.5.0'
+__version__ = '3.5.1'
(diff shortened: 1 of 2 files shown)


benvand:
Hey so couple of points here.

That `_make_slug` method could be added into `utils` instead of in the `content_loader`. Perhaps out of scope of a bugfix.

Also more abstractly, if this PR had tests you could show _exactly_ what the point of it is. This is obviously fixing some bug that a few people know about but as someone coming from the outside I can't actually tell what specific case this is designed to fix, nor whether it actually works (I'm sure it does! I just can't tell _just by looking at it_. Because I'm crap at regex-in-my-head). I guess this kind of comes under the umbrella of tests as documentation. The test, in this case, would tell me more than the code does. If you coded up a test that showed the exact problem I'd be able to tell instantly what the difference between the old and new regex is.

These aren't PR killers and if this is super urgent then by all means go ahead with the merge. Just a couple of things I thought when looking at this 💯

PR #34, opened 2017-03-13
Fix unanswered questions count

samuelhwilliams:
This PR fixes a bug in the 'answer_required' property for MultiquestionSummary, caused by a lack of knowledge about questions with followups.

Previously, MultiquestionSummaries would only do the following:
* Check all questions and return True if any of those questions require an answer.

This would cause most questions-with-dependencies to incorrectly return True if the answers given by a user did not provide answers for all questions because some of the questions were not required followups.

To fix this, a MultiquestionSummary will now do three things:
* If the multiquestion itself is marked as optional, return False.
* Iterate over all questions that have followups. For each one, if it requires answer return True. If it doesn't, check whether each of its followup are 1) required (i.e. not optional), and 2) require an answer (via its own answer_required). If yes, return True.
* Check all remaining questions (those which do not have followups, and which are not followups of other questions), and return True if any of those require an answer (via their own answer_required).

### DIFF
--- a/dmcontent/questions.py
+++ b/dmcontent/questions.py
@@ -634,10 +634,30 @@ class MultiquestionSummary(QuestionSummary, Multiquestion):
 
     @property
     def answer_required(self):
+        """Checks all sub-questions and returns true if any questions which are required, still require answers.
+        Appropriately checks/ignores followup questions based on current answers."""
         if self.get('optional'):
             return False
 
-        return any(question.answer_required for question in self.questions)
+        lookup_question_by_id = {q.id: q for q in self.questions}
+        followup_questions = set()
+
+        for question in [q for q in self.questions if q.get('followup', {})]:
+            if question.id not in followup_questions and question.answer_required:
+                return True
+
+            followup_questions.add(question.id)
+
+            answers_provided = question.value if isinstance(question.value, list) else [question.value]
+
+            for followup_id, answers_triggering_followup in question.get('followup', {}).items():
+                followup_questions.add(followup_id)
+
+                if set(answers_provided) & set(answers_triggering_followup) and \
+                        lookup_question_by_id[followup_id].answer_required:
+                    return True
+
+        return any(q.answer_required for q in self.questions if q.id not in followup_questions)
 
 
 class DynamicListSummary(MultiquestionSummary, DynamicList):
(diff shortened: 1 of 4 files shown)


samuelhwilliams:
Just writing a slightly better test for the various permutations of multiquestion summaries... Will update it soon.

dmcontent/questions.py
```diff
@@ -634,10 +634,31 @@ def value(self):
...
         if self.get('optional'):
             return False
 
-        return any(question.answer_required for question in self.questions)
+        lookup_question_by_id = {q.id: q for q in self.questions}
+        followup_questions = set()
+
+        # For each question that has followups, check the followup questions - if any require answers, the multiq does.
+        for question in self.questions:
+            answer = question.value if isinstance(question.value, list) else [question.value]
```
galund: I think it might make more sense if this variable was plural-named, seeing as its a list. Maybe 'answers' or 'selected_values' or something?

dmcontent/questions.py
```diff
@@ -634,10 +634,31 @@ def value(self):
...
+
+            followups = question.get('followup', {})
+            if followups:
+                followup_questions.add(question.id)
+
+                for followup_id, values in followups.items():
+                    followup_questions.add(followup_id)
+
+                    if set(answer) & set(values) and lookup_question_by_id[followup_id].answer_required:
```
galund: You could probably make `set(answer)` outside the loop rather than each time around. Maybe just where the `answer` var is created in the 1st place :)

dmcontent/questions.py
samuelhwilliams: Good point, thanks.

dmcontent/questions.py
samuelhwilliams: I prefer set() being in the conditional; even though it gets executed and is slightly more expensive to run, I don't think we need to worry about optimising it out and having the two set() calls in such close proximity makes it obvious what's being compared. I care more about making things easy for the developer to understand than making things easy for the computer to execute, unless it's a significant trade-off or would become a bottleneck.

risicle: This is where I rely on naming conventions e.g. `answer_set`, `values_set` - personally think it makes it just as obvious but whatever.

PR #35, opened 2017-03-27
[#141943157]Question: implement `__contains__`

risicle:
Story https://www.pivotaltracker.com/story/show/141943157

It's useful to be able to properly test whether a key is present in the `_data` - `.get()` doesn't allow you to distinguish between an absent key and a present key with a `None` value.


TheDoubleK:
Nice :+1:

PR #36, opened 2017-04-04
141882981-Add-question-type-to-content-loader-and-summary-display

benvand:
## Create handlers for processing and displaying date questions.

* Create a new `Date` class for handling date questions.
* Create a new `DateSummary` class for handling viewing dates in summary tables etc.
* Register `Date` handler in `QUESTION_TYPES` to make it auto used by all date type questions.
* Tests.

https://trello.com/c/G3J4fbtw/30-add-question-type-to-content-loader-and-summary-display

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,29 @@
 
 Records breaking changes from major version bumps
 
+## 4.0.0
+
+PR: [#36](https://github.com/alphagov/digitalmarketplace-content-loader/pull/36/files)
+
+### What changed
+
+New question type `Date` and non-backwards compatible change to `Question.unformat_data` which now returns only data
+relevant to the given question.
+
+### Example Change
+
+#### Old
+```
+        >>> question.unformat_data({"thisQuestion": 'some data', "notThisQuestion": 'other data'})
+        {"thisQuestion": 'some data changed by unformat method', "notThisQuestion": 'other data'}
+```
+#### New
+```
+        >>> question.unformat_data({"thisQuestion": 'some data', "notThisQuestion": 'other data'})
+        {"thisQuestion": 'some data changed by unformat method'}
+```
+
+
 ## 3.0.0
 
 PR: [#25](https://github.com/alphagov/digitalmarketplace-content-loader/pull/25)
@@ -12,7 +35,6 @@ Added support for `followup` questions inside multiquestions, radio questions wi
 and multiple followups for a single question. This requires changing the question YAML file
 syntax for listing followups, so the content loader is modified to support the new format
 and will only work with an updated frameworks repo.
-
 
 ### Example change
 
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -386,7 +386,7 @@ class ContentSection(object):
                 question = self.get_question(key)
                 if question:
                     # Otherwise if it is a legitimate question use the unformat method on the question.
-                    result.update(question.unformat_data(data))
+                    result.update(question.unformat_data(key, data))
                 else:
                     # Otherwise it's not a question, default to returning the k: v pair.
                     result[key] = data[key]
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -361,8 +361,7 @@ class ContentSection(object):
         }
 
     def unformat_data(self, data):
-
-        """Unpack assurance information to be used in a form
+        """Method to process form data, special assurance case or individual question level unformat.
 
         :param data: the service data as returned from the data API
         :type data: dict
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -363,8 +363,10 @@ class ContentSection(object):
     @staticmethod
     def unformat_date(key, data):
         result = {}
-        for value, field in zip(data[key].split('-'), Date.FIELDS):
-            result['-'.join([key, field])] = value
+        value = data[key]
+        if data[key]:
+            for partial_value, field in zip(value.split('-'), Date.FIELDS):
+                result['-'.join([key, field])] = partial_value
         return result
 
     def unformat_data(self, data):
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -360,15 +360,6 @@ class ContentSection(object):
             key: data[key].get('value', None)
         }
 
-    @staticmethod
-    def unformat_date(key, data):
-        result = {}
-        value = data[key]
-        if data[key]:
-            for partial_value, field in zip(value.split('-'), Date.FIELDS):
-                result['-'.join([key, field])] = partial_value
-        return result
-
     def unformat_data(self, data):
 
         """Unpack assurance information to be used in a form
@@ -389,12 +380,16 @@ class ContentSection(object):
         result = {}
         for key in data:
             if self._has_assurance(key):
+                # If it's an assurance question do the unformatting here.
                 result.update(self.unformat_assurance(key, data))
-            elif self._is_date(key):
-                result.update(self.unformat_date(key, data))
-
             else:
-                result[key] = data[key]
+                question = self.get_question(key)
+                if question:
+                    # Otherwise if it is a legitimate question use the unformat method on the question.
+                    result.update(question.unformat_data(data))
+                else:
+                    # Otherwise it's not a question, default to returning the k: v pair.
+                    result[key] = data[key]
         return result
 
     def get_question(self, field_name):
@@ -436,10 +431,6 @@ class ContentSection(object):
         """Return True if a question has an assurance component"""
         question = self.get_question(key)
         return bool(question) and question.has_assurance()
-
-    def _is_date(self, key):
-        question = self.get_question(key)
-        return question and question.is_date
 
     @property
     def has_summary_page(self):
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -11,7 +11,7 @@ from functools import partial
 from werkzeug.datastructures import ImmutableMultiDict
 
 from .errors import ContentNotFoundError, QuestionNotFoundError
-from .questions import Question, ContentQuestion
+from .questions import Question, ContentQuestion, Date
 from .messages import ContentMessage
 from .utils import TemplateField, template_all, drop_followups
 
@@ -353,7 +353,22 @@ class ContentSection(object):
 
         return errors_map
 
+    @staticmethod
+    def unformat_assurance(key, data):
+        return {
+            key + '--assurance': data[key].get('assurance', None),
+            key: data[key].get('value', None)
+        }
+
+    @staticmethod
+    def unformat_date(key, data):
+        result = {}
+        for value, field in zip(data[key].split('-'), Date.FIELDS):
+            result['-'.join([key, field])] = value
+        return result
+
     def unformat_data(self, data):
+
         """Unpack assurance information to be used in a form
 
         :param data: the service data as returned from the data API
@@ -372,8 +387,10 @@ class ContentSection(object):
         result = {}
         for key in data:
             if self._has_assurance(key):
-                result[key + '--assurance'] = data[key].get('assurance', None)
-                result[key] = data[key].get('value', None)
+                result.update(self.unformat_assurance(key, data))
+            elif self._is_date(key):
+                result.update(self.unformat_date(key, data))
+
             else:
                 result[key] = data[key]
         return result
@@ -417,6 +434,10 @@ class ContentSection(object):
         """Return True if a question has an assurance component"""
         question = self.get_question(key)
         return bool(question) and question.has_assurance()
+
+    def _is_date(self, key):
+        question = self.get_question(key)
+        return question and question.is_date
 
     @property
     def has_summary_page(self):
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -11,7 +11,7 @@ from functools import partial
 from werkzeug.datastructures import ImmutableMultiDict
 
 from .errors import ContentNotFoundError, QuestionNotFoundError
-from .questions import Question, ContentQuestion, Date
+from .questions import Question, ContentQuestion
 from .messages import ContentMessage
 from .utils import TemplateField, template_all, drop_followups
 
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -361,7 +361,7 @@ class ContentSection(object):
         }
 
     def unformat_data(self, data):
-        """Method to process form data, special assurance case or individual question level unformat.
+        """Method to process data into format for form, special assurance case or individual question level unformat.
 
         :param data: the service data as returned from the data API
         :type data: dict
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -385,7 +385,7 @@ class ContentSection(object):
                 question = self.get_question(key)
                 if question:
                     # Otherwise if it is a legitimate question use the unformat method on the question.
-                    result.update(question.unformat_data(key, data))
+                    result.update(question.unformat_data(data))
                 else:
                     # Otherwise it's not a question, default to returning the k: v pair.
                     result[key] = data[key]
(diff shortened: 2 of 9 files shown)


dmcontent/questions.py
```diff
@@ -626,6 +662,20 @@ def answer_required(self):
...
+class DateSummary(QuestionSummary):
+
+    def __init__(self, question, service_data):
+        super(DateSummary, self).__init__(question, service_data)
+        self._value = self._service_data.get(self.id, '')
+
+    @property
+    def value(self):
+        try:
+            return datetime.strptime(self._value, '%Y-%m-%d').strftime('%A %-d %B %Y')
```
idavidmcdonald: If we are formatting it into the common digital marketplace format we would ideally use the dmutils `DATE_FORMAT` instead of duplicating

tests/test_questions.py
```diff
@@ -676,6 +715,42 @@ def test_is_empty_optional_question(self):
...
+        data = {
+            "id": "example",
+            "type": "date"
+        }
+        data.update(kwargs)
+
+        return ContentQuestion(data)
+
+    def test_date_is_formatted_into_user_friendly_format(self):
+        question = self.question().summary({'example': '2016-02-18'})
```
idavidmcdonald: Is it worth having a test for where we don't include 0 padding i.e. `2`?

Either we are saving the month input directly in which case we may see data like `{'example': '2017-2-19'}` in our DB, and therefore we need to know our DateSummary is able to handle that or we only save data that includes 0 padding.

dmcontent/questions.py
```diff
@@ -544,6 +550,36 @@ def update_expected_values(options, parents, expected_set):
...
+        return True
+
+    def summary(self, service_data):
+        return DateSummary(self, service_data)
+
+    def get_data(self, form_data):
+        """Retreive the fields from the POST data (form_data).
+
+        The d, m, y should be in the post as 'questionName-day', questionName-month ...
+        Extract them and format as 'YYYY-MM-DD'.
```
idavidmcdonald: I guess technically we can't gurantee it will be 'YYYY-MM-DD' as the user input could be anything such as 'YYYY-M-D'. This is a really anal comment though...

dmcontent/content_loader.py
```diff
@@ -353,7 +353,24 @@ def get_error_messages(self, errors, question_descriptor_from="label"):
...
+        result = {}
+        value = data[key]
+        if data[key]:
+            for partial_value, field in zip(value.split('-'), Date.FIELDS):
+                result['-'.join([key, field])] = partial_value
+        return result
+
     def unformat_data(self, data):
+
         """Unpack assurance information to be used in a form
```
idavidmcdonald: If we do decide to stick with this solution then documentation for this function needs updating. Also is it worth adding a test to cover `unformat_date` and maybe calling `unformat_data` when passing a date field in?

dmcontent/questions.py
```diff
@@ -626,6 +662,20 @@ def answer_required(self):
...
+    def __init__(self, question, service_data):
+        super(DateSummary, self).__init__(question, service_data)
+        self._value = self._service_data.get(self.id, '')
+
+    @property
+    def value(self):
+        try:
+            return datetime.strptime(self._value, '%Y-%m-%d').strftime('%A %-d %B %Y')
+        except ValueError:
+            return self._value
```
idavidmcdonald: Could be nice to include a very short comment explaining that we are falling back to original date format

dmcontent/questions.py
benvand: There's a reason we we've been avoiding that. Has to do with having to add it as a dependency link in the setup.py as `digitalmarketplace-content-loader` is a module. I've added it, we can chat on Friday.

dmcontent/questions.py
```diff
@@ -182,6 +185,10 @@ def inject_brief_questions_into_boolean_list_question(self, brief):
     def has_assurance(self):
         return True if self.get('assuranceApproach') else False
 
+    @property
```
idavidmcdonald: Do we still need this?

dmcontent/questions.py
```diff
@@ -544,6 +551,43 @@ def update_expected_values(options, parents, expected_set):
         return expected_values - selected_values_set
 
 
+class Date(Question):
+    """Class used as an interface for date data between forms, backend and summary pages."""
+
+    FIELDS = ('year', 'month', 'day')
+
+    @property
```
idavidmcdonald: Ditto here, is this now redundant?

dmcontent/questions.py
```diff
@@ -544,6 +551,43 @@ def update_expected_values(options, parents, expected_set):
...
+        """Retreive the fields from the POST data (form_data).
+
+        The d, m, y should be in the post as 'questionName-day', questionName-month ...
+        Extract them and format as '\d\d\d\d-\d{1,2}-\d{1,2}'.
+        https://code.tutsplus.com/tutorials/validating-data-with-json-schema-part-1--cms-25343
+        """
+        parts = []
+        for key in self.FIELDS:
+            identifier = '-'.join([self.id, key])
+            value = form_data.get(identifier, '').replace('-', '').strip()
```
idavidmcdonald: I think we should have a test for the stripping `-` behaviour

tests/test_questions.py
```diff
@@ -168,6 +168,45 @@ def test_get_question_ids_by_type(self):
...
+            'example-month': '03',
+            'example-year': ' ',
+        }) == {'example': '-03-19'}
+
+    def test_get_data_with_blank_year(self):
+        assert self.question().get_data({
+            'example-day': '19',
+            'example-month': '03',
+            'example-year': '',
+        }) == {'example': None}
```
idavidmcdonald: Do we expect this to be `None` rather than `-03-19`?

tests/test_questions.py
idavidmcdonald: Cool, the test below now addresses this.

dmcontent/questions.py
```diff
@@ -544,6 +551,43 @@ def update_expected_values(options, parents, expected_set):
...
+            parts.append(value or '')
+
+        return {self.id: '-'.join(parts) if any(parts) else None}
+
+    def unformat_data(self, data):
```
idavidmcdonald: Do we have any tests for this method?

dmcontent/content_loader.py
```diff
@@ -353,7 +353,15 @@ def get_error_messages(self, errors, question_descriptor_from="label"):
...
+    @staticmethod
+    def unformat_assurance(key, data):
+        return {
+            key + '--assurance': data[key].get('assurance', None),
+            key: data[key].get('value', None)
+        }
+
     def unformat_data(self, data):
+
         """Unpack assurance information to be used in a form
```
idavidmcdonald: This doc block top line doesn't feel correct anymore.

dmcontent/questions.py
```diff
@@ -366,7 +369,7 @@ def get_data(self, form_data):
 
         return {self.id: questions_data}
 
-    def unformat_data(self, data):
+    def unformat_data(self, key, data):
```
idavidmcdonald: Why do we want to have two different interfaces for `unformat_data`? It felt nice that `unformat_data` had the same interface regardless of if you were calling it on a `Section` or `Question`. That pattern also mimics the `get_data` usage.

tests/test_content_loader.py
```diff
@@ -1699,11 +1699,11 @@ def test_inject_messages_into_section_and_section_summary(self):
 
 
 class TestReadYaml(object):
-    @mock.patch.object(builtins, 'open', return_value=io.StringIO(u'foo: bar'))
+    @ mock.patch('dmcontent.content_loader.open', return_value=io.StringIO(u'foo: bar'))
     def test_loading_existant_file(self, mocked_open):
         assert read_yaml('anything.yml') == {'foo': 'bar'}
 
-    @mock.patch.object(builtins, 'open', side_effect=IOError)
+    @ mock.patch('dmcontent.content_loader.open', side_effect=IOError)
```
idavidmcdonald: Should there be a space in here?

idavidmcdonald:
I'm slightly confused about what we want `unformat_data` to do.

Originally `unformat_data` would take any dictionary of data and return that data, unformatting any keys that were related to that question but everything else would be left untouched
(dynamic list questions are an example https://github.com/alphagov/digitalmarketplace-content-loader/blob/master/dmcontent/questions.py#L369)

Your new examples take a different approach of taking a dict of data but only returning data for that question e.g.
```
def unformat_data(self, data):
          return {self.id: data[self.id]}
```

Assuming consistency is desired, raises 2 questions for me:
1. Should `unformat_data` only take input related to it's own question id or take any dict of data?
2. If it should take any dict of data as input, should `unformat_data` return output only related to it's own question or return all data with it's question unformatted?

benvand:
> what we want `unformat_data` to do.
I mean `unformat_data` _can_ do whatever we want. But the reasoning behind how it functions now is that it can take all the data from a form into a question and unformat only that data that is relevant to the question.


> Should unformat_data only take input related to it's own question id or take any dict of data?
We can't know what fields are relevant to a question before the data goes in so we can't only pass in the relevant fields to be unformatted.

> If it should take any dict of data as input, should unformat_data return output only related to it's own question or return all data with it's question unformatted?
It comes down to the whole mutable dict thing. 

> Originally unformat_data would take any dictionary of data and return that data
There are 2 possibilities here. Let's say the goal of our `unformat_data` is to capitalize the value:
```
def unformat_data(self, data):
    for key in data: # iterating over a dict in python only iterates on keys
        if key == self.id:
            data[key] = data[key].capitalize()
    return data
```
or
```
# Dict comp version
def unformat_data(self, data):
    return {key: (value.capitalize() if key==self.id else value) for key, value in data.items()}

# For loop version
def unformat_data(self, data):
    d = {}
    for key in data:
        if key == self.id:
            d[key] = data[key].capitalize()
        else:
            d[key] = data[key]
```

The first mutates the given dictionary, the second creates a new one.
Mutating a dictionary inside a method is an easy way to introduce bugs that are really hard to debug. It's a bit too hidden and under the bonnet for me. I like my methods to explicitly return the values I want to use.

The second (the way it's done now)  creates a new dictionary containing all the data
>  return all data with it's question unformatted

The problem arises with the second one when we start looping over questions.
```
d={}
for question in section.questions:
    d.update(question.unformat_data(data)) # overwrites unformatted data over and over per field 
```

it's less valuable to define the new dictionary _inside_ the method because you then lose the knowledge of which fields refer to a given question.

So my way:
```
def unformat_data(self, data):
    this_questions_data_only =  {}
    for key in data:
        if key == self.id:
            this_questions_data_only(key: data[key].capitalize())
    return this_questions_data_only
```
Avoids mutation, refers only to the given question, does not return data that may already have been unformatted.
And means that we can construct the dictionary outside the method like so:
```
d={}
for question in section.questions:
    d.update(question.unformat_data(data)) # overwrites unformatted data over and over per field 
```
with no side effects.

idavidmcdonald:
Cool. I think I like your argument (thank's for taking the time to write it out clearly).

It seems sensible to go with your suggestion but we should probably make `unformat_data` on other Questions follow the same behaviour of ditching unrelated keys (https://github.com/alphagov/digitalmarketplace-content-loader/blob/master/dmcontent/questions.py#L387). It's worth taking a look first through the content loader / front end apps to make sure there is not a dependancy of our frontend to return unrelated keys.

@allait If you've got a minute, would it be possible just for a quick check on our plans for `unformat_data` to make sure we aren't missing anything?

dmcontent/questions.py
```diff
@@ -464,6 +467,9 @@ def get_question(self, field_name):
         if self.id == field_name or field_name in self.fields.values():
             return self
 
+    def unformat_data(self, data):
+        return self.get_data(data)
```
allait: I think this part is interesting. These 2 lines are confusing because `unformat_data` and `get_data` perform opposite operations, so they can't have the same implementation.

The reason we need this with the proposed implementation of `ContentSection.unformat_data` is that it delegates based on `question.get_question` and `Pricing` returns itself whenever any of it's fields are passed in. Which means that if I understand this correctly this method will get called N times for a pricing question with N fields and each time the return value will overwrite all pricing fields in the section's unformatted data dictionary. Which is not a problem because this return value is always the same.

However, this seems to suggest that we might simplify this by passing in the original field `key` to the `quesiton.unformat_data(self, data, key)` - which then allows us to use the base class implementation and return `{key: data[key]}`.

allait: Although on the other hand the same can be said about `get_data` and maybe having a similar interface between two methods makes sense.

On the other hand, they iterate over different things: `get_data` iterates over questions and `unformat_data` iterates over data keys.

benvand:
@allait _I believe_ the last 2 commits hsould address your comments. I've tried to clean up some of the code and make it clearer with comments what's going on. This has a version bump so do I need to update the changelog with the changes to unformat data on all questions?

dmcontent/questions.py
```diff
@@ -112,7 +115,7 @@ def get_error_messages(self, errors, question_descriptor_from="label"):
         return question_errors
 
     def unformat_data(self, data):
-        return data
+        return {self.id: data[self.id]}
```
idavidmcdonald: Should this use `.get` by default in case the key does not exist in data?

For some routes, we call `unformat_data` on a question directly (https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/views/briefs.py#L202) and the data might not yet exist.

dmcontent/questions.py
```diff
@@ -391,20 +394,19 @@ def unformat_data(self, data):
...
-                result[key] = data[key]
+        data = data.get(self.id, None)
+        if not data:
+            return result
+        for question in self.questions:
+            # For each question e.g. evidence-0, find if data exists for it and insert it into our result
+            root, index = question.id.split('-')
+            question_data = data[self.id][int(index)]
+            if root in question_data:
+                result.update({question.id: question_data.get(root)})
```
benvand: So one of the reasons I use update is to avoid the similar syntax when updating a list. Means that at a glace I can tell that result is a dict.

```
>>> x=[1, 2, 3, 4]
>>> y=2
>>> z=3
>>> x[y] = z
>>> x={1:2, 3:4}
>>> x[y]=z
# With the above x can be list or dict.
# With the below it can only be a dict
>>> x.update({y: z})
```

Another is that I like to use it like so especially in test fixtures:
```
def get_fake_brief(**kwargs):
    data = {
        'id':1,
        'lastChangedDate': '00:00:00.00000',
        'frameworkSlug': 'digital-outcomes-and-specialists'
        ...
    }
    data.update(kwargs)
    return data

brief1 = get_fake_brief()
brief2 = get_fake_brief(id=2)
brief3 = get_fake_brief(id=3, frameworkSlug='digital-outcomes-and-specialists-2')
```

Finally in multiple updates it avoids a for loop:

```
my_nonsense_dict = {}
my_updates = {'foo': 'bar', 'baz': 'quux'}
for key in my_updates:
    my_nonsense_dict[key] = my_updates[key]


my_nonsense_dict = {}
my_updates = {'foo': 'bar', 'baz': 'quux'}
my_nonsense_dict.update(my_updates)
```

So yeah it's kind of personal preference but founded on the fact that if I want to do the above things _and_ keep my code consistent then I have to use `.update` everywhere.

benvand:
OK, should be good to go @idavidmcdonald 
Version bumped and changelog updated

CHANGELOG.md
```diff
@@ -2,6 +2,29 @@
 
 Records breaking changes from major version bumps
 
+## 4.0.0
+
+PR: [#36](https://github.com/alphagov/digitalmarketplace-content-loader/pull/36/files)
+
+### What changed
+
+New question type `Date` and non-backwards compatible change to `Question.unformat_data` which not returns only data
```
idavidmcdonald: typo "not" should be "now" I assume

idavidmcdonald:
There is a single typo that might be nice to fix but apart from that awesome work!
👍 👍 👍 👍 👍 👍 👍
💯 💯 💯 💯 💯 💯 💯 
💯 💯 💯 💯 💯 💯 💯
👍 👍 👍 👍 👍 👍 👍

PR #37, opened 2017-04-10
BUGFIX-failing-YAML-test

benvand:
`tests/test_content_loader.py::TestReadYaml::test_loading_existant_file`
The above was failing on master when utils was added as a dependency, there's been a refactor recently but not sure what caused it tbh.

Fixed mock to point at instance open that is used by the function being
tested.

### To reproduce

```
git clone https://github.com/alphagov/digitalmarketplace-content-loader.git
cd digitalmarketplace-content-loader
vitrualenv venv
source venv/bin/activate
echo -e '\ngit+https://github.com/alphagov/digitalmarketplace-utils.git@25.0.1#egg=digitalmarketplace-utils==25.0.1' >> requirements.txt
pip install -r requirements_for_test.txt
py.test
```

### DIFF
--- a/tests/test_content_loader.py
+++ b/tests/test_content_loader.py
@@ -1699,11 +1699,11 @@ class TestContentSection(object):
 
 
 class TestReadYaml(object):
-    @mock.patch.object(builtins, 'open', return_value=io.StringIO(u'foo: bar'))
+    @mock.patch('dmcontent.content_loader.open', return_value=io.StringIO(u'foo: bar'))
     def test_loading_existant_file(self, mocked_open):
         assert read_yaml('anything.yml') == {'foo': 'bar'}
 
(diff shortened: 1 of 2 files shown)


tests/test_content_loader.py
```diff
@@ -1699,11 +1699,11 @@ def test_inject_messages_into_section_and_section_summary(self):
 
 
 class TestReadYaml(object):
-    @mock.patch.object(builtins, 'open', return_value=io.StringIO(u'foo: bar'))
+    @ mock.patch('dmcontent.content_loader.open', return_value=io.StringIO(u'foo: bar'))
```
pcraig3: Is there meant to be a space between `@` and `mock`?

benvand:
@pcraig3 @idavidmcdonald Updated

PR #38, opened 2017-04-21
Tiny utility function for getting checkbox option values.

galund:
- will also be used by script in the frameworks API (for generating Elasticsearch mappings).

https://trello.com/c/etYh34Zs/341-fixing-software-parent-categories


idavidmcdonald:
👍

PR #39, opened 2017-04-24
Fix missing dm-utils dependency

galund:
Add dm-utils dependency to setup.py, as well as requirements.txt:

 - required so that apps dependent on the Content Loader fail clearly if they do not already have dm-utils in _their_ requirements.txt.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,6 @@ setup(
         'Werkzeug==0.11.9',
         'inflection==0.3.1',
         'six==1.10.0',
-        'digitalmarketplace-utils>=25.0.1',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -25,6 +25,7 @@
         'PyYAML==3.11',
         'Werkzeug==0.11.9',
         'inflection==0.3.1',
-        'six==1.10.0'
+        'six==1.10.0',
+        'digitalmarketplace-utils>=25.1.0',
```
galund: Have updated this again cc @benvand ;-)

PR #42, opened 2017-05-19
Add filter label values to question summary

TheDoubleK:
Adds a `filter_value` property to question summaries that will replace values with their `filter_label` from content if it exists, or fall back to the usual value otherwise.

### DIFF
--- a/dmcontent/questions.py
+++ b/dmcontent/questions.py
@@ -661,6 +661,17 @@ class QuestionSummary(Question):
         return value
 
     @property
+    def filter_value(self):
+        # For options where we want to show different text on the service page than when the question was asked
+        options = self.get('options')
+        value = self._service_data.get(self.id, '')
+        if options and value:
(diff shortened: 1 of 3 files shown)


dmcontent/questions.py
```diff
@@ -776,6 +788,28 @@ def value(self):
...
+    def filter_value(self):
+        # Display values for options where we want to show different text ('filter_label') than the ususal 'label'
+        options = self.get('options')
+        value = self._service_data.get(self.id, '')
+        if options and value:
+            value = list(value)  # Make a copy of the list to avoid mutating the underlying list data
+            for i, v in enumerate(value):
+                for option in options:
+                    if 'filter_label' in option and 'value' in option and v == option['value']:
+                        value[i] = option['filter_label']
```
galund: Can we just do `value[i] = option['filter_label'] or option['label']` and then we only need one branch?

PR #44, opened 2017-05-26
Make validation messages into TemplateFields

pcraig3:
Some of our declaration questions want to be able to reference other questions, so that we can say things like 
```if you answered question 32, you have to answer this one```. 

The syntax they use for this is a bit peculiar. It looks like 
```if you answered question [[questionId]], you have to answer this one```.

I'm trying to get rid of this, in favour of something like
```if you answered question {{ content.get_question('questionId').number }}, you have to answer this one.```

Since v2 of the Content Loader, we can [use jinja to render `TemplateField`s](https://github.com/alphagov/digitalmarketplace-content-loader/blob/master/CHANGELOG.md#200).  The fields we allow to be templated are whitelisted -- for example, `question` and `hint` accept jinja templating, but `type` and `id` don't.  

Some of the values we want to replace in the declaration questions are in validation messages (ie, [mitigatingFactors2.yml](https://github.com/alphagov/digitalmarketplace-frameworks/blob/fe77c4134ad56c5847807b5ddc1c73046d60a3d2/frameworks/digital-outcomes-and-specialists-2/questions/declaration/mitigatingFactors2.yml#L7)) but validation messages aren't whitelisted as `TemplateField`s.

This pull request makes the `message` value of `validation` items into `TemplateField`s.

***

**Note**

If this pull request goes in, there's a couple of other things to follow up with:
- updating the README in the frameworks repo with all `TemplateField`s (ie, [like this](https://github.com/alphagov/digitalmarketplace-frameworks/tree/pc-update-readme-with-more-template-fields))
- removing the `question_references` stuff entirely in the supplier app (ie, [more or less like this](https://github.com/alphagov/digitalmarketplace-supplier-frontend/tree/pc-kill-the-double-square-brackets))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,7 +80,7 @@ These are the fields that are turned into TemplateFields when the YAML files are
 
 | ContentSection                    | Question                                           |
 |-----------------------------------|----------------------------------------------------|
-| [`name`, `description`](https://github.com/alphagov/digitalmarketplace-content-loader/blob/474d9adce0f422700cbf2dfc8815a7503ab368bc/dmcontent/content_loader.py#L123) | [`question`, `name`, `question_advice`, `hint`](https://github.com/alphagov/digitalmarketplace-content-loader/blob/474d9adce0f422700cbf2dfc8815a7503ab368bc/dmcontent/questions.py#L10) |
+| [`name`, `description`](https://github.com/alphagov/digitalmarketplace-content-loader/blob/474d9adce0f422700cbf2dfc8815a7503ab368bc/dmcontent/content_loader.py#L123) | [`question`, `name`, `question_advice`, `hint`](https://github.com/alphagov/digitalmarketplace-content-loader/blob/474d9adce0f422700cbf2dfc8815a7503ab368bc/dmcontent/questions.py#L10), and [`options[i].description`, and `validations[i].message`](https://github.com/alphagov/digitalmarketplace-content-loader/blob/29b1be3571558312a4b6271c06d04b65b44dd607/dmcontent/questions.py#L14) <sub>since v4.4.2</sub> |
 
 In addition, all string fields (including fields in lists and nested dictionaries) in content
 messages are turned into TemplateFields.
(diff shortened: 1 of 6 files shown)


samuelhwilliams:
`[[questionId]]` feels more friendly to content writers, if we're still working on the assumption and desire for the frameworks repo to be accessible to non-developers. The suggested replacement, `{{ content.get_question('questionId').number }}`, is now really mingling code and content, moving us farther away from any goal to have this be maintained primarily by designers. If we're accepting that, fine, but if we still have it in the back of our heads that this should be accessible to non-developers, it would be nice to keep anything 'magical' (content-wise) as simple as possible. Double square brackets, while magical, are at least fairly simple to understand.

pcraig3:
Yeah, point taken. It is harder to use.

I guess it depends what our priorities are.  Right now, we have two ways of swapping out values: one that uses double square brackets and one that doesn’t. This removes one way of doing value substitution so that we’re consistently using the “new” Jinja-rendering way: it’s more flexible for us, and it makes our codebase more consistent, but it’s also much less readable for non-coders.

I think what I would argue is that most of these declaration questions are copy+pasted anyway for new frameworks, so if the square brackets notation isn't being used very often when writing new questions, then we don't have much of a practical issue.  (for example, `appropriateTradeRegistersNumber` is the same across [DOS](https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists/questions/declaration/appropriateTradeRegistersNumber.yml) and [DOS2](https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists-2/questions/declaration/appropriateTradeRegistersNumber.yml), and `mitigatingFactors2` _has_ been updated from [G8](https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/g-cloud-8/questions/declaration/mitigatingFactors2.yml) to [G9](https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/g-cloud-9/questions/declaration/mitigatingFactors2.yml), but only to make the content work better with markdown.)

I would be happy to rewrite the G9 and DOS2 questions to this format to facilitate copy+pasting sometime down the road if that's a compromise.

To sum up: it _is_ harder to use the new format but I don't think it's a big problem in practice. However, if we do think that it is, then fair enough and I'll look for another solution to this.

pcraig3:
@samuelhwilliams, you've been outvoted.

PR #46, opened 2017-09-27
Flake8

benvand:
[flake8-report.txt](https://github.com/alphagov/digitalmarketplace-content-loader/files/1337429/flake8-report.txt)

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '4.4.2'
+__version__ = '4.5.0'
--- a/dmcontent/questions.py
+++ b/dmcontent/questions.py
@@ -430,7 +430,7 @@ class DynamicList(Multiquestion):
             question_errors[input_name] = {
                 'input_name': input_name,
                 'question': getattr(question, question_descriptor_from),
-                'message':  question.get_error_message(error['error']),
+                'message': question.get_error_message(error['error']),
             }
 
         return question_errors
@@ -666,10 +666,9 @@ class QuestionSummary(Question):
         options = self.get('options')
         value = self._service_data.get(self.id, '')
         if options and value:
-            for option in options:
-                if ('filter_label' in option and 'value' in option and (
-                            option.get('value') or option.get('label')) == value):
-                    return option['filter_label']
+            for opt in options:
+                if 'filter_label' in opt and 'value' in opt and any(value == opt.get(i) for i in ['value', 'label']):
+                    return opt['filter_label']
         return self.value
 
     @property
--- a/tests/test_questions.py
+++ b/tests/test_questions.py
@@ -934,23 +934,20 @@ class TestCheckboxTreeSummary(QuestionSummaryTest):
     def test_value(self):
         question = self.question().summary({'example': ['val1.1', 'val2.2', 'val4']})
         assert question.value == [
-                {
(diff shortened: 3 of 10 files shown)


dmcontent/__init__.py
```diff
@@ -2,4 +2,4 @@
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '4.4.2'
+__version__ = '4.5.0'
```
TheDoubleK: There's no new functionality added to content loader here, so I'd probably argue for a patch version bump.

dmcontent/questions.py
```diff
@@ -667,9 +667,9 @@ def filter_value(self):
         value = self._service_data.get(self.id, '')
         if options and value:
             for option in options:
-                if ('filter_label' in option and 'value' in option and (
-                            option.get('value') or option.get('label')) == value):
-                    return option['filter_label']
+                if 'filter_label' in option and 'value' in option:
```
TheDoubleK: Splitting up the conditional and nesting `ifs` here as a fix for "continuation line over-indented for hanging indent" doesn't feel like strictly a Flake8 fix. You could just have deleted some spaces for that.
Personally I prefer the single `if` as it was, but it's marginal and either works I guess ¯\\_(ツ)_/¯

tests/test_questions.py
```diff
@@ -934,23 +934,18 @@ def question(self, **kwargs):
...
-                    "value": "val2",
-                    "options": [
-                        {"label": "Option 2.2", "value": "val2.2", "options": []},
-                    ]
-                },
-
-                {"label": "Option 4", "value": "val4", "options": []},
-            ]
+            {
+                "label": "Option 1", "value": "val1", "options": [
```
TheDoubleK: I think this was easier to follow with each key on its own line.

dmcontent/__init__.py
benvand: I'm going to stick with Minor because of the decision I came to on the utils PR (same/ similar changes are included here):

https://github.com/alphagov/digitalmarketplace-utils/pull/338/files
https://govuk.slack.com/archives/C04AXAXEM/p1506433136000301
>>> I'm going to make a call on this. It is my opinion that, technically, according to the semver spec, I've changed the interface for these packages, but not strictly broken backwards compatibility for _documented_ api functionality (both by removing imports and changing the Makefile commands). I'm going to release Flake8 changes with `Makefile` and import changes as minor version bumps, in line with the semver spec.



http://semver.org/

tests/test_questions.py
benvand: Fixed, best of both worlds 👍

dmcontent/questions.py
benvand: The single if will look like this otherwise 
```
        if options and value:
            for option in options:
                if (
                    'filter_label' in option 
                    and 'value' in option
                    and (option.get('value') or option.get('label')) == value
                ):
                        return option['filter_label']
```

Also I think splitting out the separate actions (ie 'checking the keys are in the dict' and 'checking the values of keys in the dict') creates a nice context boundary. 

For me it reads like:

if key one and key two are there
    if value two or value three match `value`
        do the thing

I like how i can disregard the first if statement while i'm working out what the second bit does (which feels a bit gnarly to me).

Something like this would be nicer 
`if 'filter_label' in option and 'value' in option and value in [option.get('value'), option.get('label')]:`

But that is 122 characters.

benvand: Nailed it.

benvand: Yep, nailed it.

PR #48, opened 2017-10-18
Add ability to set pricing question decimal place restriction

idavidmcdonald:
https://trello.com/c/PHnolEgP/16-change-money-format-on-awarding-a-brief-to-be-pounds-and-pence

Users had previously been able to say that a contract was worth £8.105. This should not be allowed. Instead we only allow 0 or 2 decimal places e.g £8 or £8.10 to be stored in the database. This new content loader property can be used by the frameworks repo to know when to use the stricter format for `pricing` questions and generate the new JSON schemas for them.

In addition, we make the validation permissive so if a user enters '10.0' or '10.', it will be transformed into '10.00' and their price will validate rather then forcing them to type their answer in the exact format we require. 

permissive2

See frameworks repo PR for further details - https://github.com/alphagov/digitalmarketplace-frameworks/pull/477

### DIFF
--- a/dmcontent/questions.py
+++ b/dmcontent/questions.py
@@ -459,6 +459,7 @@ class Pricing(Question):
     def __init__(self, data, *args, **kwargs):
         super(Pricing, self).__init__(data, *args, **kwargs)
         self.fields = data['fields']
+        self.decimal_place_restriction = data.get('decimal_place_restriction', False)
 
     def summary(self, service_data):
         return PricingSummary(self, service_data)
--- a/dmcontent/questions.py
+++ b/dmcontent/questions.py
@@ -1,5 +1,6 @@
 from collections import OrderedDict, defaultdict
 from datetime import datetime
+import re
 
 from dmutils.formats import DATE_FORMAT, DISPLAY_DATE_FORMAT
 
@@ -459,6 +460,8 @@ class Pricing(Question):
     def __init__(self, data, *args, **kwargs):
         super(Pricing, self).__init__(data, *args, **kwargs)
         self.fields = data['fields']
+
+        # True if we are restricting to an integer or a 2dp value (representing pounds and optionally pence)
         self.decimal_place_restriction = data.get('decimal_place_restriction', False)
 
     def summary(self, service_data):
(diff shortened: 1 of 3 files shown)


dmcontent/questions.py
```diff
@@ -459,6 +459,7 @@ class Pricing(Question):
     def __init__(self, data, *args, **kwargs):
         super(Pricing, self).__init__(data, *args, **kwargs)
         self.fields = data['fields']
+        self.decimal_place_restriction = True if data.get('decimal_place_restriction') else False
```
samuelhwilliams: Is this cleaner just as `data.get('decimal_price_restriction', False)`?

idavidmcdonald: Yes, good point. Will change.

dmcontent/questions.py
```diff
@@ -478,7 +481,19 @@ def get_data(self, form_data):
...
+        question_data = {key: form_data[key] or None for key in self.fields.values() if key in form_data}
+
+        if self.decimal_place_restriction:
+            # Permissive validation to transform value to 2dp if value contains a `.` but no decimal values or is 1dp
+            pattern_1dp = re.compile(r".*(\.)(?:\d)$")
+            for key, value in question_data.items():
+                if value[-1] == '.':
+                    question_data[key] = value + "00"
+
+                if pattern_1dp.search(value):
```
samuelhwilliams: We don't need a compiled pattern here; we could just inline it as `if re.match(r".*(\.)(?:\d)$", value):`

Doesn't _really_ matter much either way, of course...

tests/test_questions.py
```diff
@@ -307,6 +307,27 @@ def test_pricing_question_decimal_place_restriction_is_false_by_default(self):
...
+        assert question.get_data({'priceMin': '10'}) == {'priceMin': '10'}
+
+    def test_get_data_with_decimal_place_restriction_does_not_edit_price_if_more_than_2dp(self):
+        question = self.question(decimal_place_restriction=True)
+        assert question.get_data({'priceMin': '10.123'}) == {'priceMin': '10.123'}
+
+    @pytest.mark.parametrize("price", ["11", "11.", "11.2", "11.23", "11.234"])
+    def test_get_data_does_not_edit_price_if_decimal_restriction_is_false(self, price):
+        question = self.question()
+        assert question.get_data({'priceMin': price}) == {'priceMin': price}
```
samuelhwilliams: These tests make me think. Clearly there's a lot of repetition, which is good to take out where possible, but I like how the test names clearly document the logic behind each assertion, which would be lost if we parametrized this lot into one test. However, maybe the data would provide enough detail for the correct conclusion to be drawn?

I wonder.

```

@pytest.mark.parametrize('decimal_place_restriction, submitted_value, anticipated_output_value',
                         (
                             (True,  '10.',    '10.00',),
                             (True,  '10.1',   '10.10',),
                             (True,  '10',     '10',),
                             (True,  '10.123', '10.123',),

                             (False, '11',     '11',),
                             (False, '11.',    '11.',),
                             (False, '11.2',   '11.2',),
                             (False, '11.23',  '11.23',),
                             (False, '11.234', '11.234',),
                         )    
```

Questionable.

PR #52, opened 2017-11-27
Fix multi question ordering

katstevens:
Trello ticket: https://trello.com/c/77L1kwqj/28-2-list-of-errors-in-header-in-incorrect-order-for-brief-weighting-question

The content loader was returning errors for multi-part questions in alphabetical order. This fix returns them in the same order as the questions (see below for example).

~~I opted for a major version bump as this is likely to break any tests that assert error messages as a list - if I'm being overcautious I'm happy to downgrade to a minor bump!~~ Going for a minor bump as discussed below.

#### Old
```
>>> question = ContentQuestion({"questions": [{"id": "technicalCompetence", ...}, {"id": "culturalFit", ...}])
>>> question.get_error_messages({"technicalCompetence": "an error happened", "culturalFit": "another error happened"})
OrderedDict([
      ("culturalFit", {"message": "an error happened", ...}),
      ("technicalCompetence", {"message": "another error happened", ...}),
])
```
#### New
```
>>> question = ContentQuestion({"questions": [{"id": "technicalCompetence", ...}, {"id": "culturalFit", ...}])
>>> question.get_error_messages({"technicalCompetence": "an error happened", "culturalFit": "another error happened"})
OrderedDict([
      ("technicalCompetence", {"message": "another error happened", ...}),
      ("culturalFit", {"message": "an error happened", ...}),
])
```

### DIFF
--- a/tests/test_questions.py
+++ b/tests/test_questions.py
@@ -472,6 +472,50 @@ class TestMultiquestion(QuestionTest):
             ('follow', 'a')
         ])) == {'lead': ['no', 'maybe not'], 'follow': None}
 
+    def test_get_error_messages_orders_by_question_order(self):
+        question = self.question(
+            questions=[
+                {
+                    "id": "technicalCompetence",
+                    "type": "text",
+                    "label": "Technical Competence",
+                },
+                {
+                    "id": "price",
+                    "type": "text",
+                    "label": "Price",
+                },
+                {
+                    "id": "culturalFit",
+                    "type": "number",
+                    "label": "Cultural Fit",
+                }
+            ]
+        ).filter(self.context())
+
+        multi_question_form_errors = {
+            "culturalFit": "answer_required",
+            "price": "not_a_number",
+            "technicalCompetence": "answer_required",
+        }
+        assert question.get_error_messages(multi_question_form_errors) == OrderedDict([
+            ('technicalCompetence', {
+                'input_name': 'technicalCompetence',
+                'message': 'You need to answer this question.',
+                'question': 'Technical Competence'
+            }),
+            ('price', {
+                'input_name': 'price',
+                'message': 'There was a problem with the answer to this question.',
+                'question': 'Price'
+            }),
+            ('culturalFit', {
+                'input_name': 'culturalFit',
+                'message': 'You need to answer this question.',
+                'question': 'Cultural Fit'
+            })
+        ])
+
 
 class TestDynamicListQuestion(QuestionTest):
     default_context = {'context': {'field': ['First Need', 'Second Need', 'Third Need', 'Fourth need']}}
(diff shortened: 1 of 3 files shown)


tests/test_questions.py
```diff
@@ -472,6 +474,27 @@ def test_nested_checkboxes_question_followup_get_data(self):
...
 
+    def test_get_error_messages_orders_by_question_order(self):
+        question = self.question().filter(self.context())
+
+        multi_question_form_errors = OrderedDict([
+            ('example3', 'not_a_number'),
+            ('example2', 'answer_required')
+        ])
+
+        assert question.get_error_messages(multi_question_form_errors) == OrderedDict([
```
idavidmcdonald: Please correct me if I'm mistaken but in this test, `get_error_messages` looks like it returns error messages in the same order of the questions as defined in `self.question()` however that is also the alphabetical order. Therefore, how do we know that our new code is reordering them based on order and not keeping the existing behaviour?

tests/test_questions.py
```diff
@@ -472,6 +474,27 @@ def test_nested_checkboxes_question_followup_get_data(self):
             ('follow', 'a')
         ])) == {'lead': ['no', 'maybe not'], 'follow': None}
 
+    def test_get_error_messages_orders_by_question_order(self):
+        question = self.question().filter(self.context())
+
+        multi_question_form_errors = OrderedDict([
```
idavidmcdonald: Is the order of the form errors supposed to be important for this test? If so, could you explain why please (sorry, it's been a while since I've looked at the content loader).

katstevens: Yeah, I was trying to show that even if the errors came through to the method in a different order, the new method would sort by question order. However as you point out, the question order in the setup is alphabetical anyway :woman_facepalming: I'll fix that up for this test now.

idavidmcdonald:
@TheDoubleK It looks like we were previously asking for errors to be sorted into alphabetical order. I don't suppose you happen to know why this might have been the case? Do we rely on alphabetical ordering of error messages elsewhere? It doesn't feel like something product would have suggested

TheDoubleK:
I don't get how this is a breaking change?  Existing code should still run just fine, only have the errors in a sensible order?

katstevens:
@TheDoubleK yeah, I wasn't sure whether potentially failing tests would count as a breaking change. Happy to do a minor bump instead.

TheDoubleK:
@idavidmcdonald I've got no idea.  It looks like Multiquestion never had a `get_error_messages` of its own before, so maybe was inheriting the alphabetical behaviour from something where it made more sense?

TheDoubleK:
Yeah I reckon minor bump for this.  No application code will need to change to pull in the new version - breaking test data is fine (i.e. not a breaking change).

katstevens:
@TheDoubleK it's inheriting from the base `Question.get_error_messages`, which presumably is getting all the errors for a single field. If the errors are displayed next to each other (above that field) then alphabetical seems like a sane way of sorting them. It's just not so good when we have the header with different sub-sections to jump to.

PR #54, opened 2018-01-11
Add 'try_load_manifest' generic method for apps.

samuelhwilliams:
## Summary
Add a util method to try loading a manifest in an application. The tests might be controversial as they literally just check what the method is doing internally, but that feels like the right thing to do in this case I think the contract this method makes is simply that another method is called - what that method itself does shouldn't be tested here, I don't think.

Bump version to 4.9.0

## Related PR
https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/68

## Ticket
https://trello.com/c/tnBdllTV/156-briefs-frontend-has-crashed-on-preview-wont-come-back-up

### DIFF
--- a/dmcontent/utils.py
+++ b/dmcontent/utils.py
@@ -5,6 +5,7 @@ from markdown import markdown
 from six import string_types
 
 from dmutils.jinja2_environment import DMSandboxedEnvironment
+from dmcontent.errors import ContentNotFoundError
 
 from .errors import ContentTemplateError
 
@@ -98,3 +99,13 @@ def get_option_value(option):
     :return: string value to be persisted
     """
     return option.get('value') or option['label']
+
+
+def try_load_manifest(content_loader, application, data, question_set, manifest):
+    try:
+        content_loader.load_manifest(data['slug'], question_set, manifest)
+
+    except ContentNotFoundError:
+        application.logger.info(
+            "Could not load {}.{} manifest for {}".format(question_set, manifest, data['slug'])
+        )
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,10 +1,11 @@
 # -*- coding: utf-8 -*-
 
+import mock
 import pytest
 from jinja2 import Environment, Markup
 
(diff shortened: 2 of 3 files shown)


dmcontent/utils.py
```diff
@@ -98,3 +99,14 @@ def get_option_value(option):
...
     """
     return option.get('value') or option['label']
+
+
+def try_load_manifest(content_loader, application, data, question_set, manifest):
+    try:
+        content_loader.load_manifest(data['slug'], question_set, manifest)
+
+    except ContentNotFoundError:
+        # Not all frameworks have this, so no need to panic (e.g. G-Cloud 4, G-Cloud 5, G-Cloud-6)
```
TheDoubleK: The comment doesn't make sense in this context.

tests/test_utils.py
```diff
@@ -160,3 +161,45 @@ def test_get_option_value():
...
+
+    def test_content_loader_asked_to_load_manifest(self):
+        try_load_manifest(self.content_loader_mock, self.application_mock,
+                          TestTryLoadManifest.FRAMEWORK_DATA,
+                          TestTryLoadManifest.QUESTION_SET,
+                          TestTryLoadManifest.MANIFEST
+                          )
+
+        assert self.content_loader_mock.load_manifest.called is True
+        self.content_loader_mock.load_manifest.assert_called_with(
```
TheDoubleK: Should this maybe be `assert_called_once_with`?

Or even (I know @katstevens prefers this way) 
```
self.content_loader_mock.load_manifest.call_args_list == [
    mock.call(TestTryLoadManifest.FRAMEWORK_DATA['slug'], 
              TestTryLoadManifest.QUESTION_SET,
              TestTryLoadManifest.MANIFEST)
]
```

dmcontent/utils.py
samuelhwilliams: Stupid comment...

tests/test_utils.py
samuelhwilliams: I'm OK with the latter way. I'll tweak.

PR #55, opened 2018-01-25
Straw man for moving question references into the content loader

idavidmcdonald:
STRAWMAN

Appears to work in principle so we would no longer need this
in the template layer.

Few issues that I haven't looked at yet include
- setting question data properties without using private properties
- Template fields (making sure this doesn't screw things up)
- How to decide which fields to run question referencing through
- How to potentially make this behaviour optional
- Proper tests etc
- The other tests that I've broken

### DIFF
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -35,8 +35,32 @@ class ContentManifest(object):
                 question_index += 1
                 question.number = question_index
 
+        for section in self.sections:
+            for question in section.questions:
+                if question.get("hint"):
+                    question._data["hint"] = self.question_references(question["hint"])
+
+
     def __iter__(self):
         return self.sections.__iter__()
+
+    def question_references(self, data):
+        """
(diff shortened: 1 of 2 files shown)


dmcontent/content_loader.py
```diff
@@ -35,9 +35,33 @@ def __init__(self, sections):
                 question_index += 1
                 question.number = question_index
 
+        for section in self.sections:
+            for question in section.questions:
+                if question.get("hint"):
+                    question._data["hint"] = self.question_references(question["hint"])
```
galund: Accessing a field on question like this feels a bit weird. I definitely think this is the right kind of approach, but potentially calling this out into a separate method might make what's going on here more self-documenting.

So for example (and I might have got the name a bit wrong) here we could do:

`question.updateHintReferences(self)`

and then all of the logic here and in your `question_references` function might be able to go onto the question class?

idavidmcdonald:
Closing this as deciding it is too much effort for the moment.

PR #56, opened 2018-04-12
Load arbitrary framework metadata

samuelhwilliams:
## Summary
We need to be able to copy services from one framework to another during
applications to save suppliers time. In order to do this, we need to
define which previous framework has services that are eligible for
copying, and which questions are suitable. This information appears to
be data related to the framework, so we are storing it in dm-frameworks
as metadata about a given framework (in this case, G-Cloud 10). In order
for a frontend to have that information, we need to be able to load it.
The `messages` folder of dm-frameworks stores files that contain
key-value pairs, and this is the same kind of information format we want
for our metadata, so this creates a stripped-down version of the same
which works on the `metadata` folder of a framework. A ContentMetadata
instance does not parse the data it reads and therefore does not support
templated fields; this seems like a reasonable conclusion to make as we
are simply storing static data that describes how the framework
interacts with others.

## Related PR
https://github.com/alphagov/digitalmarketplace-frameworks/pull/515

### DIFF
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -463,6 +463,9 @@ class ContentLoader(object):
     >>>
     >>> # get a message
     >>> loader.get_message('framework-1', 'homepage_sidebar', 'in_review')
+    >>>
+    >>> # get some metadata
+    >>> loader.get_message('framework-1', 'copy_services', 'source_framework')
 
     """
     def __init__(self, content_path):
--- a/dmcontent/content_loader.py
+++ b/dmcontent/content_loader.py
@@ -13,6 +13,7 @@ from werkzeug.datastructures import ImmutableMultiDict
 from .errors import ContentNotFoundError, QuestionNotFoundError
 from .questions import Question, ContentQuestion
 from .messages import ContentMessage
+from .metadata import ContentMetadata
 from .utils import TemplateField, template_all, drop_followups
 
 
@@ -454,6 +455,9 @@ class ContentLoader(object):
     >>> # preload messages
     >>> loader.load_messages('framework-1', ['homepage_sidebar', 'dashboard'])
     >>>
+    >>> # preload metadata
+    >>> loader.load_metadata('framework-1', ['copy_services'])
+    >>>
     >>> # get a manifest
     >>> loader.get_manifest('framework-1', 'manifest-1')
     >>>
@@ -465,6 +469,8 @@ class ContentLoader(object):
         self.content_path = content_path
         self._content = defaultdict(dict)
         self._messages = defaultdict(dict)
+        self._metadata = defaultdict(dict)
+
         # A defaultdict that defaults to a defaultdict of dicts
         self._questions = defaultdict(partial(defaultdict, dict))
 
@@ -573,6 +579,41 @@ class ContentLoader(object):
     def _load_message(self, framework_slug, message_name):
         return template_all(read_yaml(self._message_path(framework_slug, message_name)))
 
+    def get_metadata(self, framework_slug, block, key=None):
+        """
+        `block` corresponds to
+          - a file in the frameworks directory
+
+        `key` and `sub_key` are used to look up a specific metadata:
+          - the framework which services can be copied from
+        """
+        if block not in self._metadata[framework_slug]:
+            raise ContentNotFoundError(
+                "Message file at {} not loaded".format(self._metadata_path(framework_slug, block))
+            )
+
+        metadata = ContentMetadata(self._metadata[framework_slug][block])
+
+        if key is None:
+            return metadata
+        else:
+            return metadata.get(key, None)
+
+    def load_metadata(self, framework_slug, blocks):
+        if not isinstance(blocks, list):
+            raise TypeError('Content blocks must be a list')
+
+        for block in blocks:
+            try:
+                self._metadata[framework_slug][block] = self._load_metadata(framework_slug, block)
+            except IOError:
+                raise ContentNotFoundError(
+                    "No metadata file at {}".format(self._metadata_path(framework_slug, block))
+                )
+
+    def _load_metadata(self, framework_slug, metadata_name):
+        return read_yaml(self._metadata_path(framework_slug, metadata_name))
+
     def _root_path(self, framework_slug):
         return os.path.join(self.content_path, 'frameworks', framework_slug)
 
@@ -581,6 +622,9 @@ class ContentLoader(object):
 
     def _message_path(self, framework_slug, message):
         return os.path.join(self._root_path(framework_slug), 'messages', '{}.yml'.format(message))
+
+    def _metadata_path(self, framework_slug, metadata):
+        return os.path.join(self._root_path(framework_slug), 'metadata', '{}.yml'.format(metadata))
 
     def _load_nested_questions(self, framework_slug, question_set, section_or_question):
         if 'questions' in section_or_question:
--- a/dmcontent/metadata.py
+++ b/dmcontent/metadata.py
@@ -1,4 +1,4 @@
-class ContentMetadata(object):
+class ContentMetadata:
     """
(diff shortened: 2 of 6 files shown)


dmcontent/content_loader.py
```diff
@@ -454,6 +455,9 @@ class ContentLoader(object):
     >>> # preload messages
     >>> loader.load_messages('framework-1', ['homepage_sidebar', 'dashboard'])
     >>>
+    >>> # preload metadata
+    >>> loader.load_metadata('framework-1', ['copying-services'])
+    >>>
```
Wynndow: Might be worth adding a `loader.get_metatdata()` example as per manifest and message below.

dmcontent/content_loader.py
```diff
@@ -582,6 +623,9 @@ def _questions_path(self, framework_slug, question_set):
     def _message_path(self, framework_slug, message):
         return os.path.join(self._root_path(framework_slug), 'messages', '{}.yml'.format(message))
 
+    def _metadata_path(self, framework_slug, metadata):
```
Wynndow: These `_<something>_path` methods are all really similar. Could they be combined and parameterised?

dmcontent/metadata.py
```diff
@@ -0,0 +1,40 @@
+class ContentMetadata(object):
```
Wynndow: As all the apps are now Python3, do we need to inherit from `object`?

tests/test_content_loader.py
```diff
@@ -2033,13 +2029,83 @@ def test_caching_of_messages(self, mock_read_yaml):
...
+    def test_get_metadata_with_key(self, mock_read_yaml):
+        mock_read_yaml.return_value = {
+            'field_one': 'value_one',
+            'field_two': 'value_two',
+        }
+        metadata = ContentLoader('content/')
+        metadata.load_metadata('g-cloud-7', ['index'])
+
+        assert metadata.get_metadata('g-cloud-7', 'index', 'field_one') == 'value_one'
+
```
Wynndow: Coud test the error case of `get_metadata` when the block isn't found.

tests/test_content_loader.py
```diff
@@ -2033,13 +2029,83 @@ def test_caching_of_messages(self, mock_read_yaml):
...
+    def test_load_metadata_wraps_nested_fields_with_template_field(self, mock_read_yaml):
```
Wynndow: Not sure I understand what template field the refers to? Is the test to ensure that subkeys in the yaml are loaded correctly if they're a list for example?

tests/test_content_loader.py
```diff
@@ -2033,13 +2029,83 @@ def test_caching_of_messages(self, mock_read_yaml):
...
+        })
+
+        assert metadata.get_metadata('g-cloud-7', 'copy_services').source_framework == 'g-cloud-6'
+
+    def test_load_metadata_argument_types(self, mock_read_yaml):
+        mock_read_yaml.return_value = {}
+        metadata = ContentLoader('content/')
+
+        with pytest.raises(TypeError):
+            metadata.load_metadata('g-cloud-7', 'index')  # blocks argument must be a list
```
Wynndow: Could check the error message instead of having the comment.

tests/test_content_loader.py
```diff
@@ -2033,13 +2029,83 @@ def test_caching_of_messages(self, mock_read_yaml):
...
+
+    def test_get_metadata_must_preload(self, mock_read_yaml):
+        mock_read_yaml.return_value = {}
+        metadata = ContentLoader('content/')
+        metadata.load_metadata('g-cloud-8', ['index'])
```
Wynndow: This is confusing. You're testing that it hasn't loaded data that you haven't loaded. Where does the "preload" come in?

tests/test_content_loader.py
```diff
@@ -2033,13 +2029,83 @@ def test_caching_of_messages(self, mock_read_yaml):
...
+            metadata.get_metadata('g-cloud-8', 'copy_services')
+            mock_read_yaml.assert_not_called()
+
+    def test_caching_of_metadata(self, mock_read_yaml):
+        metadata = ContentLoader('content/')
+        metadata.load_metadata('g-cloud-7', ['copy_services'])
+        metadata.get_metadata('g-cloud-7', 'copy_services').get('source_framework')
+        metadata.get_metadata('g-cloud-7', 'copy_services').get('source_framework')
+
+        mock_read_yaml.assert_called_once_with('content/frameworks/g-cloud-7/metadata/copy_services.yml')
```
Wynndow: Not sure what this is testing. You've only called `load_metadata` once, so the mock will only be called once? `get_metadata` doesn't touch it? Also, where does the caching happen?

tests/test_metadata.py
```diff
@@ -0,0 +1,46 @@
+from dmcontent.metadata import ContentMetadata
+from dmcontent.utils import TemplateField
+
+
+class TestContentMetadata(object):
+    def test_content_metadata_field(self):
```
Wynndow: To be more explicit, you could call this `test_content_metadata___get_attr__`. Or something.

Wynndow: Could also test that it raises an attribute error if not found.

dmcontent/metadata.py
samuelhwilliams: I was getting a weird flake8/pep8 thing when I dropped that... So I left it in. 🙈 It wanted me to define an `__init__` method, but it would be empty. I guess I can do it anyway.

samuelhwilliams: Ok not getting the problem now. So yes, I'll remove the inherit. 👍

dmcontent/content_loader.py
samuelhwilliams: 👍

dmcontent/content_loader.py
samuelhwilliams: They are similar, but I think abstracting this would make some simple methods more confusing than they need to be. The duplication is fine here IMO.

tests/test_content_loader.py
samuelhwilliams: Basically all of these tests are copied from `test_messages`, but yeah you're right on the whole that we could probably do more. I'll review.

tests/test_content_loader.py
samuelhwilliams: TemplateField is used to allow custom context for messages, so that a message `Hello, {{ name }}` stored in the frameworks repo could be displayed on the frontend as eg `Hello, Chris`.

For metadata we don't want this dynamic context, so it's just asserting that it isn't available.

tests/test_content_loader.py
samuelhwilliams: Done.

tests/test_content_loader.py
samuelhwilliams: Yeah, I'll drop that initial `load_metadata` call.

Wynndow: Is this testing that you need to load the metadata before you can get it?

tests/test_content_loader.py
samuelhwilliams: Yeah this is clearly a completely pointless test. I'll delete it from ContentMetadata and ContentMessage.

Wynndow:
OK, a weird one, but the content load currently specifies `Jinja2==2.8`. Supplier frontend specifies 2.10. If I use pip to install your version of the content loader with the `-e` flag, it breaks at least one of our templates. It seems to be to do with the way that imports are scoped. If `summary-table.html` is imported within a block, it's not passed on to blocks defined within than block. With 2.10 it seems to. Which is weird.

I'm guessing this is a quirk with installing it editably. When we install the dependencies in the usual way, I suspect that we're left with 2.10. Doing it editably overwrites them. Either way, it might be worth bumping the version in here.

samuelhwilliams:
Upgraded to Jinja2==2.10 👍

PR #60, opened 2018-09-17
Update requirements

lfdebrux:
This PR includes a bunch of house-cleaning stuff:

  - Use venv instead of virtualenv
  - Pin to version ranges instead of specific versions
  - Upgrade dmutils a few minor versions
  - Fix an annoying deprecation warning
  - Change our library for uploading to coveralls to one that is currently maintained

This contributes to closing https://trello.com/c/WoFsBnLE/

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '5.0.0'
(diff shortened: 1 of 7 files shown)


dmcontent/__init__.py
```diff
@@ -2,4 +2,4 @@
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '5.0.0'
+__version__ = '5.1.0'
```
katstevens: 👍

PR #65, opened 2019-02-04
default.nix: ensure venv is created relative to default.nix, not cwd

risicle:
uninteresting to non-nixers


risicle:
Interesting travis is failing for this - it must also be failing for our master.

risicle:
That seems to have fixed it

PR #67, opened 2019-03-20
Update digitalmarketplace-utils from 46.2.0 to 47.1.1

lfdebrux:


katstevens:
This will need its own version bump, however, is this causing a conflict somewhere? If not I probably wouldn't bother updating it yet...

lfdebrux:
> This will need its own version bump, however, is this causing a conflict somewhere? If not I probably wouldn't bother updating it yet...

It may cause a conflict when (if) we want to update digitalmarketplace-utils for other apps, but it's not urgent.

PR #70, opened 2019-06-17
TemplateField: make this deepcopy-able

risicle:
https://trello.com/c/RW2tzjbM

Being able to deepcopy a full `ContentLoader` tree gives us a fast way of generating independent `ContentLoaders` for each local context. This approach cheats slightly my transforming the way we hold a reference to a jinja `Template` to be immutable, and therefore not actually _requiring_ a copy. This also means we can _share_ the jinja `Templates` associated with `TemplateField`s, which seem to be the most memory hungry and construction-time hungry parts of a `ContentLoader` tree.

I'm probably being a bit over-cautious in making this a major version bump. I don't think anything touches the internal `template` attribute of a `TemplateField`.

### DIFF
--- a/dmcontent/utils.py
+++ b/dmcontent/utils.py
@@ -14,6 +14,24 @@ from .errors import ContentTemplateError
 template_environment = DMSandboxedEnvironment(autoescape=True, undefined=StrictUndefined)
 
 
+class _ImmutableTemplateProxy:
+    """
+        A heavily abbreviated proxy for a jinja Template that should help ensure (effective) immutability and
+        threadsafety
+    """
+    __slots__ = ("render",)
+
+    def __init__(self, template):
+        # instead of keeping a reference to the template on the instance where external code could be e.g. inadvertantly
+        # calling mutating methods on it, seal it in the closure of a function which itself will _become_ our render
+        # "method".
+        self.render = lambda *args, **kwargs: template.render(*args, **kwargs)
+
+    def __deepcopy__(self, memo):
+        # we're (effectively) immutable.
+        return self
+
+
 class TemplateField(object):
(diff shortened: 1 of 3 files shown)


dmcontent/utils.py
```diff
@@ -9,6 +9,28 @@
...
+    __slots__ = ("render",)
+
+    def __init__(self, template):
+        # instead of keeping a reference to the template on the instance where external code could be e.g. inadvertantly
+        # calling mutating methods on it, seal it in the closure of a function which itself will _become_ our render
+        # "method".
+        self.render = lambda *args, **kwargs: template.render(*args, **kwargs)
+
+    def __deepcopy__(self, memo):
+        # we're immutable.
```
lfdebrux: Are we actually immutable? What's to stop someone setting `render`?

lfdebrux: Might be worth changing the comment just so people don't get the wrong idea.

risicle: Well, there's "immutable" and _immutable_. It's extremely tricky to make a pure-python object actually _immutable_ - `tuple` et al manage it because they're implemented in c, and even then they're not deep-immutable (a `dict` in a `tuple` is still mutable). I _could_ add some more chicanery to make a few more things impossible, but then I would open myself to the accusation that I'm introducing something that's overly-complex and confusing.

It's more about what anyone's ever likely to do without it being obvious that they're being naughty - e.g. it's _possible_ to modify `app.config` from within a flask request, but it's fairly obvious that that would be a bad idea and people using `app.config` don't tend to worry about the possibility that it could change during a request.

risicle:
Might change it to "effectively immutable".

PR #71, opened 2019-10-21
Upgrade PyYAML to 5.x

katstevens:
https://trello.com/c/51brdJKx/1201-upgrade-pyyaml-to-5x

I've been cautious and made this a major bump, in case anyone was thinking of using this with an older version of Python.

We are careful to use `yaml.safeload()` instead of the vulnerable `yaml.load()`, but this version of PyYAML deprecates the latter completely so it's no longer even an option 😸

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Records breaking changes from major version bumps
 
 ## 7.0.0
 
-PR [TBC]()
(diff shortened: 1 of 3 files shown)


katstevens:
@risicle Yep - will pull it into a content-heavy app like Brief Responses FE and run the functional tests locally.

PR #72, opened 2019-11-12
remove max utils version restriction

risicle:
https://trello.com/c/wIs7vHwR

~I'm not convinced of the usefulness of this version restriction tbh...~ gone

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ setup(
         'Markdown<3.0.0,>=2.6.7',
         'PyYAML<6.0,>=5.1.2',
         'inflection<1.0.0,>=0.3.1',
-        'digitalmarketplace-utils<49.0.0,>=44.0.1',
+        'digitalmarketplace-utils>=44.0.1',
     ],
     python_requires="==3.6.*",
 )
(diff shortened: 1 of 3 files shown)


lfdebrux:
> I'm not convinced of the usefulness of this version restriction tbh...

Let's remove it then! ;)

risicle:
I'm ok with that - can anyone think of any gotchas?

setup.py
```diff
@@ -25,7 +25,6 @@
         'Markdown<3.0.0,>=2.6.7',
         'PyYAML<6.0,>=5.1.2',
         'inflection<1.0.0,>=0.3.1',
-        'digitalmarketplace-utils<49.0.0,>=44.0.1',
```
lfdebrux: Does this library not require digitalmarketplace-utils at all?

katstevens: It uses the date formatters (and the `DMSandboxedEnvironment` for handling `TemplateField` types).

lfdebrux: I think this should be

```suggestion
        'digitalmarketplace-utils>=44.0.1',
```

PR #76, opened 2020-01-14
performance improvements, largely through addition of "inplace_allowed" options

risicle:
https://trello.com/c/ooTuJDCw

On my machine,when using the `inplace_allowed` options, this runs the test workload mentioned in https://trello.com/c/RsEEo4So in about 0.59x the time current `master` does.

The bulk of the speedup here comes from new `inplace_allowed` options for `.filter()` and `.summary()`. I'd previously observed that a lot of the time spent in the content loader during normal usage patterns is spent performing deep copy upon deep copy of itself. Currently `.filter()` and `.summary()` calls will return an altered deep-ish copy of themselves ("deep-ish" because it's worse in some places than others). An extremely common usage pattern of the loader however is to call e.g.

```
content_loader.get_manifest("foo").filter(service).summary(service)
```

with each of `get_manifest`, `filter`, and `summary` performing a deep copy of their own. This is very wasteful, because in this sort of call chain, the existing objects are being discarded. So really there's no need to be making sure changes only get applied to the copy that's being returned. Once `get_manifest` has given us its deep copy, we should be satisfied that we have an independent copy of this data that we can mutate as we like.

So the new `inplace_allowed` flag on `.filter()` and `.summary()` lifts this restriction, allowing these methods to operate on (and return) _themselves_. It can be used as and when the caller deems it suitable, but I suspect this will actually be _many_ places in our code.

Note that this argument is called `_allowed` - it doesn't guarantee that the implementation being called will actually mutate itself - there are many cases where it makes no sense to do so. So when using this flag, the original object being called should be discarded and only the return value used, because there's guarantee of which state the object will now be in.

I've gone a bit overboard on the testing for this and turned almost every test which calls either `.filter()` or `.summary()` into a parametrized test, testing both modes.

Other smaller improvements contribute less but are still measurable. They mostly come from intra-function optimizations in very hot paths. Commit comments tell you more...

### DIFF
--- a/dmcontent/questions.py
+++ b/dmcontent/questions.py
@@ -191,16 +191,13 @@ class Question(object):
         return [self.id] if type in [self.type, None] else []
 
     def get(self, key, default=None):
-        try:
-            return getattr(self, key)
-        except AttributeError:
-            return default
+        return getattr(self, key, default)
 
     def __getattr__(self, key):
-        try:
-            field = self._data[key]
-        except KeyError:
+        if key not in self._data:
             raise AttributeError(key)
+
+        field = self._data[key]
 
         if isinstance(field, TemplateField):
             return field.render(self._context)
--- a/dmcontent/questions.py
+++ b/dmcontent/questions.py
@@ -28,7 +28,7 @@ class Question(object):
         if not self._should_be_shown(context):
             return None
 
-        question = ContentQuestion(self._data, number=self.number)
+        question = self if inplace_allowed else ContentQuestion(self._data, number=self.number)
         question._context = context
 
         return question
--- a/dmcontent/questions.py
+++ b/dmcontent/questions.py
@@ -185,7 +185,7 @@ class Question(object):
             self.boolean_list_questions = brief.get(self.id, [])
 
     def has_assurance(self):
-        return True if self.get('assuranceApproach') else False
+        return bool(self.get('assuranceApproach'))
 
     def get_question_ids(self, type=None):
         return [self.id] if type in [self.type, None] else []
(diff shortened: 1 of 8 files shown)


dmcontent/questions.py
```diff
@@ -742,24 +745,29 @@ def answer_required(self):
             return False
 
         lookup_question_by_id = {q.id: q for q in self.questions}
-        followup_questions = set()
+        followups_optionals = set()
```
katstevens: Aren't we excluding optional sub-questions a couple of lines above? Why has this set been renamed?

risicle: No, that excludes optional multiquestions, not subquestions. I changed the name because I realized that questions also get added to this set if they're not `.answer_required`, they don't have to be mentioned as another question's followup. Though I now realize that a question _not_ being `.answer_required` doesn't _necessarily_ make it "optional" - it could be already completed, so `followups_optionals` isn't technically accurate either.

risicle: Ok better now.

risicle:
I've now tried this on preview in the `admin` and `supplier` frontends with all appropriate locations switched to `inplace_allowed=True` and the FTs pass.

dmcontent/questions.py
```diff
@@ -736,30 +739,43 @@ def value(self):
...
-
-                if set(answers_provided) & set(answers_triggering_followup) and \
+                if answers_provided_set.intersection(answers_triggering_followup) and \
                         lookup_question_by_id[followup_id].answer_required:
                     return True
 
-        return any(q.answer_required for q in self.questions if q.id not in followup_questions)
+                # ignorable because it's listed as a followup to a question that hasn't been triggered or is not
+                # `.answer_required`
+                ignorable_ids.add(followup_id)
```
katstevens: Cool, this makes sense to me 👍

PR #77, opened 2020-02-28
markdown: add GOVUKFrontendExtension to apply govuk-frontend classes to generated elements

risicle:
https://trello.com/c/scaNu2Aq

What do people think? To me, doing this seems less scary than keeping half of our content fields as fully-escaped html.

~Haven't tried this in a frontend yet.~

### DIFF
--- a/dmcontent/markdown.py
+++ b/dmcontent/markdown.py
@@ -1,0 +1,44 @@
+from typing import Mapping
+from xml.etree.ElementTree import Element
+
+from markdown.extensions import Extension
+from markdown.treeprocessors import Treeprocessor
+
+
+class _ClassAddingTreeprocessor(Treeprocessor):
+    _mapping: Mapping[str, str]
+
+    def __init__(self, mapping: Mapping[str, str]):
+        self._mapping = mapping
+
+    def run(self, root: Element):
+        self._run(root)
+
+    def _run(self, element: Element):
+        # recurse
+        for child in element:
+            self._run(child)
+
+        # mutate each affected element in-place
+        if element.tag in self._mapping:
+            orig_class = element.attrib.get("class", "")
+            element.set("class", (orig_class + " " if orig_class else "") + self._mapping[element.tag])
+
+
+class GOVUKFrontendExtension(Extension):
+    """
+    Markdown extension adding govuk-frontend classes to generated paragraphs and lists (note - will not affect
+    manual html)
+    """
+    GOVUK_ELEMENT_CLASSES = {
+        "p": "govuk-body",
+        "ul": "govuk-list govuk-list--bullet",
+        "ol": "govuk-list govuk-list--number",
+    }
+
+    def extendMarkdown(self, md, md_globals):
+        # NOTE the interface for registering extensions is much improved after Markdown 3.0.0. TODO upgrade.
+        md.registerExtension(self)
+        self.processor = _ClassAddingTreeprocessor(self.GOVUK_ELEMENT_CLASSES)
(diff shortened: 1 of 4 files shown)


dmcontent/markdown.py
```diff
@@ -0,0 +1,42 @@
...
+    Markdown extension adding govuk-frontend classes to generated paragraphs and lists (note - will not affect
+    manual html)
+    """
+    def extendMarkdown(self, md, md_globals):
+        # NOTE the interface for registering extensions is much improved after Markdown 3.0.0. TODO upgrade.
+        md.registerExtension(self)
+        self.processor = _ClassAddingTreeprocessor({
+            "p": "govuk-body",
+            "ul": "govuk-list govuk-list--bullet",
+            "ol": "govuk-list govuk-list--number",
```
katstevens: Could we pull this out separately so it's super easy to append additional classes to later if needed?

dmcontent/markdown.py
```diff
@@ -0,0 +1,42 @@
...
+        # mutate each affected element in-place
+        if element.tag in self._mapping:
+            orig_class = element.attrib.get("class", "")
+            element.set("class", (orig_class + " " if orig_class else "") + self._mapping[element.tag])
+
+
+class GovUkFrontendExtension(Extension):
+    """
+    Markdown extension adding govuk-frontend classes to generated paragraphs and lists (note - will not affect
+    manual html)
```
katstevens: What happens here with escaping? Could we add a test?

dmcontent/markdown.py
risicle: Yes, there's even a mechanism for doing that built into the extension business but I decided not to wrestle with that now seeing as I didn't even know where the config should be kept.

dmcontent/markdown.py
risicle: Oh I thought I included an ampersand in the test text but I didn't. I'll do that.

risicle: Done

dmcontent/markdown.py
risicle: Done

katstevens:
Cool 👍

risicle:
Does this look ok to people? It's actually only a little bit wider spaced than it was before
Spectacle h32487

risicle:
Before
Spectacle n32487

risicle:
Though it does fix the wrong-sized list text font problem with narrow screens
Spectacle k32487

risicle:
Whereas without the new classes...
Spectacle T32487

katstevens:
Yep, spacing looks fine to me (sizing fix obviously a win as well).

dmcontent/markdown.py
```diff
@@ -0,0 +1,44 @@
...
+        for child in element:
+            self._run(child)
+
+        # mutate each affected element in-place
+        if element.tag in self._mapping:
+            orig_class = element.attrib.get("class", "")
+            element.set("class", (orig_class + " " if orig_class else "") + self._mapping[element.tag])
+
+
+class GovUkFrontendExtension(Extension):
```
lfdebrux: This should probably be named `GOVUKFrontendExtension` to be consistent with [govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/899264cd89559d31f0a790682254d94ecf66418b/package/govuk/common.js)

risicle: :+1:

katstevens:
Re-running the Py3.8 Travis check, which stalled trying to build its environment.

PR #79, opened 2020-05-29
Add html module to format questions for display

domoscargin:
https://trello.com/c/CXGExCLO/55-3-replace-summary-table-with-summary-list-on-g-cloud-service-page-and-dos-opportunity-page

We want to be able to easily display users answers to questions using
the [summary list component](https://design-system.service.gov.uk/components/summary-list/) from the GOV.UK Design System.

To be able to do this we need to turn a collection of QuestionSummary
objects into an list of row objects usable with the `govukSummaryList`
macro.

To achieve this, we have added the `html` module which formats `QuestionSummary` values into HTML and arranges them into a list that can be passed straight into a `{{ govukSummaryList 
 rows={transformed_data} }}` macro.

### DIFF
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -1,0 +1,436 @@
+from textwrap import dedent
+
+import jinja2
+import pytest
+from jinja2 import Markup
+
+from dmcontent.content_loader import ContentManifest
+from dmcontent.html import to_html, to_summary_list_rows
+
+
+@pytest.fixture
+def content_summary():
+    content = ContentManifest(
+        [
+            {
+                "slug": "first_section",
+                "name": "First Section",
+                "questions": [
+                    {
+                        "id": "myBooleanTrue",
+                        "type": "boolean",
+                        "label": "Boolean question with answer True",
+                    },
+                    {
+                        "id": "myBooleanFalse",
+                        "type": "boolean",
+                        "label": "Boolean question with answer False",
+                    },
+                    {
+                        "id": "myBooleanList",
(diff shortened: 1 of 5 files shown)


tests/test_html.py
```diff
@@ -0,0 +1,437 @@
...
+@pytest.fixture
+def content_summary():
+    content = ContentManifest(
+        [
+            {
+                "slug": "first_section",
+                "name": "First Section",
+                "questions": [
+                    {
+                        "id": "boolean_true",
```
katstevens, line 20: Are the `id`s in a real content manifest going to be snake case or camel case? It's a small thing really, but when you're scanning down a piece of code, casing helps differentiate between the python variables, the type strings that are always going to be the same, and test data that will vary from manifest to manifest. 

I'd probably go for something like `{"id": "myBooleanQ1", ... ` instead?

domoscargin: camelCaseConfirmed - did some renaming.

tests/test_html.py
```diff
@@ -0,0 +1,437 @@
...
+
+    assert template.render(html=html) == expected
+
+
+def test_to_summary_list_rows(content_summary):
+    questions = content_summary.sections[0].questions
+    summary_list_rows = to_summary_list_rows(questions)
+
+    assert summary_list_rows
+    assert len(summary_list_rows)
```
katstevens, line 403: We may as well be explicit here - `assert len(summary_list_rows) == 22` (and then we can remove the assertion above it).

PR #81, opened 2020-06-05
Fix bug(s) in dmcontent.html.text_to_html

lfdebrux:


tests/test_html.py
```diff
@@ -175,6 +175,29 @@ def content_summary():
     )
 
 
+def test_text_to_html_takes_one_string_value_and_returns_markup():
+    assert text_to_html("") == Markup("")
+    assert text_to_html("Hello World") == Markup("Hello World")
+
```
domoscargin, line 181:
```suggestion
assert text_to_html("hello world") == Markup("hello world")
```

To check it doesn't capitalise first if no kwargs passed?

tests/test_html.py
```diff
@@ -175,6 +175,29 @@ def content_summary():
...
+def test_text_to_html_can_capitalize_first_letter():
+    assert text_to_html("hello world", capitalize_first=True) == "Hello world"
+    assert text_to_html("hello world", capitalize_first=False) == "hello world"
+
+
+def test_text_to_html_can_format_links():
+    assert text_to_html("https://gov.uk", format_links=True) == \
+        Markup('<a href="https://gov.uk" class="app-break-link" rel="external">https://gov.uk</a>')
+    assert text_to_html("https://gov.uk", format_links=False) == "https://gov.uk"
```
domoscargin, line 191: Similar check here (passing no kwargs) just to guard against link formatting being done automatically somewhere in the future?

tests/test_html.py
lfdebrux: Good idea

tests/test_html.py
```diff
@@ -175,6 +175,29 @@ def content_summary():
...
+
+
+def test_text_to_html_can_preserve_line_breaks():
+    assert text_to_html("First line.\r\nSecond line.", preserve_line_breaks=True) == \
+        Markup("First line.<br>Second line.")
+    assert text_to_html("First line.\r\nSecond line.", preserve_line_breaks=False) == \
+        Markup("First line.\r\nSecond line.")
+
```
domoscargin, line 199: And something here? I'm not sure if passing `preserve_line_breaks=False` is the same thing as passing nothing in Python.

lfdebrux: Good question! It depends on how you specify the keyword arguments; if they have a default such as `def text_to_html(preserve_line_breaks=False)` then passing nothing would be the same as passing `False`.

However, if they don't have a default, which is the case here because the definition is `def text_to_html(**kwargs)`, then passing nothing means that the argument does not exist at all. Python doesn't have a value of undefined like JavaScript does; trying to access an undefined variable results in an exception.

lfdebrux: In fact in general using `**kwargs` like I am here is to be avoided. I should probably rewrite it :D

PR #86, opened 2020-06-11
Add new tab link option to to_summary_list_rows

domoscargin:
On the [Briefs FE preview Brief page](https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/13a289f2387697f3139158bc2b16b46771b1742f/app/templates/buyers/preview_brief_source.html#L123), we use the property `open_links_in_new_tab`, which we hadn't accounted for.

This adds that property to the HTML script.

Required for https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/321

https://trello.com/c/BpDhnaYa/93-1-preview-dos-opportunity-before-publishing

### DIFF
--- a/dmcontent/html.py
+++ b/dmcontent/html.py
@@ -23,7 +23,11 @@ import dmutils.filters as filters
 
 
 def to_html(
-    question_summary, *, capitalize_first: bool = False, format_links: bool = False
+    question_summary,
+    *,
+    capitalize_first: bool = False,
+    format_links: bool = False,
+    open_links_in_new_tab: bool = False
 ) -> Markup:
     """Format the value of a QuestionSummary as HTML.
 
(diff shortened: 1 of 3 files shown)


dmcontent/html.py
```diff
@@ -100,7 +106,10 @@ def text_to_html(
         value = filters.capitalize_first(value)
 
     if format_links is True:
-        value = filters.format_links(value)
+        if open_links_in_new_tab is True:
+            value = filters.format_links(value, open_links_in_new_tab)
+        else:
+            value = filters.format_links(value)
```
lfdebrux, line 112: We can simplify this to 
```suggestion
        value = filters.format_links(value, open_links_in_new_tab)
```

dmcontent/html.py
```diff
@@ -88,6 +93,7 @@ def text_to_html(
         capitalize_first=False,
         format_links=False,
         preserve_line_breaks=False,
+        open_links_in_new_tab=False,
```
lfdebrux, line 96: Can we add the new option to the docstring below?

dmcontent/html.py
domoscargin: Ha, so we can!

PR #88, opened 2020-06-22
Update digitalmarketplace-utils from 52.2.0 to 52.6.0

lfdebrux:


requirements.txt
```diff
@@ -1,3 +1,3 @@
 -e .
 # DON'T FORGET TO UPDATE setup.py !!
```
katstevens, line 2: I think we forgot to update `setup.py` 😹

lfdebrux: 😅 woops 🤦

lfdebrux: On the other hand, the library still works with version 52.2.0 of dmutils, it's just the tests that change, so maybe the version in setup.py could stay the same? What do you think?

katstevens: I'd rather keep them consistent if we can.

PR #90, opened 2020-08-11
Add dmcontent.govuk_frontend for creating govuk-frontend components from Question objects

lfdebrux:
This is my contribution to https://trello.com/c/YXsBOrDK/110-3-replace-text-boxes-with-govuk-frontend-text-input-component-in-create-a-dos-opportunity-journey. It tries to suggest a way forward for creating govuk-frontend forms from content loader objects.

The basic approach is to move almost all of the logic into Python code. I've put this logic into the content loader repo in this PR, so it can be more easily tested and type-checked (and perhaps one day more easily deleted!). This contrasts to the approach that digitalmarketplace-frontend-toolkit took; all of the formatting and dispatch logic that used to be in the macros `digitalmarketplace-frontend-toolkit/templates/forms` should (in my opinion) live in this module. The benefit of this is easier testing, as well as (in my opinion) increased readability.

The one thing that this code can't and in my opinion shouldn't do is calling the Jinja macros themselves. This does mean we need some extra Jinja magic in our frontend apps; The [briefs frontend PR that goes with this PR](https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/337) I think shows how this could be done.

The main function in this module and in this PR is `dmcontent.govuk_frontend.from_question`. My design intent with this is that it should encapsulate the dispatch logic i.e. it should be the main entry point if you have a Question object and just want to get parameters out. The idea is to make things simpler for developers and have sensible defaults. Let me know if this sounds like a bad idea.

On the subject of sensible defaults, there are quite a few things that could be configurable that aren't right now; I decided to err on the side of "you're not going to need it" and not make the functions very flexible. If that needs to change it should be straightforward however, and I think the benefit of the simpler code is worth it for now.

There's a lot in this PR, so please do read and comment with questions, or leave a thumbs up if it all makes sense to you.

### _params function

As noted in the commit message and docstring, this function is an implementation detail and should not be used outside of this module. It is going have to change in future, especially so we can handle labels and legends/fieldsets differently, but there wasn't a need to put that logic in to fulfil the definition of done of the textbox ticket.

I made the decision to split the parameter logic out now rather than waiting because I wanted to show how we can test the logic at a fine-grained level for the internal function and then do higher-level functional tests for the exported functions that get used.

### Labels as page headings

In this PR we always use labels as page headings in the one question per page pattern. That obviously won't work for all of the pages on the Digital Marketplace, however it works for the pages I cared about for this ticket.

I decided not to add the functionality to configure it to leave open different possible ways to configure (or not) in future. For instance, it might be that we can use some property of the question to decide whether the label should be the page heading or not. When we get to multiquestions and more than one question per page this will be something that will need to be considered further, but I think we should park it for now.

### Question advice

This PR also suggests an approach for how to handle the question advice (see https://trello.com/c/RAQa597F/203-spike-figure-out-what-to-do-about-question-advice for background on this).

~~I found the easiest thing to do was to put the question advice in the hint parameter, wrapped with a `div` so it can be styled differently.~~

After talking about the question advice with @domoscargin, we've agreed on a slightly different design for dmcontent.govuk_frontend that puts more responsibility on the caller to put the form together correctly, but gives a better user experience.

See [discussion below](https://github.com/alphagov/digitalmarketplace-content-loader/pull/90#discussion_r468572289) for more details.

### Snapshot testing

To try and keep the tests of functions in govuk_frontend both comprehensive and readable, I've used snapshot testing with `syrupy` to move the expected values into a separate file.

The benefit of this is that the tests are more readable and easier to update (just run `venv/bin/pytest --update-snapshots`), and if the test fails we get very pretty diffs highlighting the change in question.

The downside is that the expected value is now outside of the test file, which means a bit less transparency.

If people hate this I can remove the commit(s) in question and leave the expected value in the tests.

### DIFF
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -11,7 +11,18 @@ from dmutils.forms.errors import govuk_error
 from dmcontent.questions import Question
 
 
-__all__ = []
+__all__ = ["govuk_input"]
+
+
+def govuk_input(
+    question: Question, data: Optional[dict] = None, errors: Optional[dict] = None
+) -> dict:
+    """Create govukInput macro parameters from a text question"""
+
+    params = _params(question, data, errors)
+    params["classes"] = "app-text-input--height-compatible"
+
+    return params
 
 
 def _params(
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -1,0 +1,92 @@
+"""
+Create forms to answer questions using govuk-frontend macros.
+"""
+
+from typing import Optional
+
+from jinja2 import Markup, escape
+
+from dmutils.forms.errors import govuk_error
+
+from dmcontent.questions import Question
+
+
+__all__ = []
+
+
+def _params(
+    question: Question, data: Optional[dict] = None, errors: Optional[dict] = None
+) -> dict:
+    """Common parameters for govuk-frontent components
+
+    The GOV.UK Design System macro library govuk-frontend has a consistent set
+    of parameters that are used across almost all of its component macros.
+
+    This function abstracts out those common parameters to hopefully simplify
+    creating parameter sets for specific macros.
+
+    *This function should however be considered an implementation detail of
+    this module, and you should avoid using it outside of this file (hence the
+    underscore).*
+
+    The parameters handled by this function include:
+
+        - errorMessage (optional)
+        - hint (optional)
+        - id
+        - label
+        - name
+        - value (optional)
+
+    :returns: A dictionary with parameters that are generally useful for
+              govuk-frontend component macros
+    """
+    params = {
+        "id": f"input-{question.id}",
+        "name": question.id,
+    }
+
+    label_text = question.question
+    if question.is_optional:
+        # GOV.UK Design System says
+        # > mark the labels of optional fields with '(optional)'
+        label_text += " (optional)"
+
+    params["label"] = {
+        # Style the label as a page heading, following the
+        # GOV.UK Design System question pages pattern at
+        # https://design-system.service.gov.uk/patterns/question-pages/
+        "classes": "govuk-label--l",
+        "isPageHeading": True,
+
+        "text": label_text,
+    }
+
+    hint_html = Markup()
+    if question.get("question_advice"):
+        # Put the question advice inside the hint, wrapped in a div
+        # We add the class .app-hint--text so the question advice
+        # can be styled like a normal paragraph with the following Sass
+        #
+        #     .app-hint--text {
+        #       @extend %govuk-body--m
+        #     }
+        hint_html += Markup('<div class="app-hint--text">\n')
+        hint_html += escape(question.question_advice)
+        hint_html += Markup("\n</div>")
+        if question.get("hint"):
+            hint_html += "\n"
+
+    if question.get("hint"):
+        hint_html += escape(question.hint)
+
+    if hint_html:
+        params["hint"] = {"html": hint_html}
+
+    if data and question.id in data:
+        params["value"] = data[question.id]
+
+    if errors and question.id in errors:
+        params["errorMessage"] = govuk_error(errors[question.id])["errorMessage"]
+
+    return params
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -1,5 +1,11 @@
 """
 Create forms to answer questions using govuk-frontend macros.
+
+The main function in this module is `from_question`. It should be possible to
+do everything you might want to do just by calling `from_question` with a
+content loader Question.
+
+Read the docstring for `from_question` for more detail on how this works.
 """
 
 from typing import Optional
@@ -11,7 +17,55 @@ from dmutils.forms.errors import govuk_error
 from dmcontent.questions import Question
 
 
-__all__ = ["govuk_input"]
+__all__ = ["from_question", "govuk_input"]
+
+
+def from_question(
+    question: Question, data: Optional[dict] = None, errors: Optional[dict] = None
+) -> Optional[dict]:
+    """Create parameters object for govuk-frontend macros from a question
+
+    `from_question` aims to solve the developer need of
+
+        Given a content loader Question
+        I want to create a form element using the GOV.UK Design System
+        So that the user gets a good experience
+
+    in a way that requires the developer to know as little as possible about
+    the Question object in question.
+
+    `from_question` takes a Question and returns a dict containing the name of
+    the govuk-frontend macro to call and the parameters to call it with.
+    Calling the macro with the parameters is left to the app developer.
+
+        >>> from dmcontent import Question
+        >>> from_question(Question({'id': 'q1', 'type': 'text', 'question': ...})
+        {'macro_name': 'govukInput', 'params': {...}}
+
+    A little bit of Jinja magic is required for this;  you need a table of
+    macro names to macros in a Jinja template:
+
+        {% set govuk_forms = {
+            'govukInput': govukInput,
+            'govukRadios': govukRadios,
+            ...
+        } %}
+
+        {% set form = from_question(question) %}
+
+        {{ govuk_forms[question.macro_name](question.params) }}
+
+    :param question: A Question or QuestionSummary
+    :param data: A dict that may contain the answer for question
+    :param errors: A dict which may contain an error message for the question
+
+    :returns: The name of the macro and the parameters in a dict, or None
+              if we don't know how to handle this type of question
+    """
+    if question.type == "text":
+        return {"macro_name": "govukInput", "params": govuk_input(question, data, errors)}
+    else:
+        return None
 
 
 def govuk_input(
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -10,14 +10,12 @@ Read the docstring for `from_question` for more detail on how this works.
 
 from typing import Optional
 
-from jinja2 import Markup, escape
-
 from dmutils.forms.errors import govuk_error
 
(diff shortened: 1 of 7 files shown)


dmcontent/govuk_frontend.py
```diff
@@ -0,0 +1,152 @@
...
+    params = {
+        "id": f"input-{question.id}",
+        "name": question.id,
+    }
+
+    label_text = question.question
+    if question.is_optional:
+        # GOV.UK Design System says
+        # > mark the labels of optional fields with '(optional)'
+        label_text += " (optional)"
```
domoscargin, line 113: 🙌

dmcontent/govuk_frontend.py
```diff
@@ -0,0 +1,92 @@
...
+from dmutils.forms.errors import govuk_error
+
+from dmcontent.questions import Question
+
+
+__all__ = []
+
+
+def _params(
+    question: Question, data: Optional[dict] = None, errors: Optional[dict] = None
```
katstevens, line 78: 🏆 for `typing`

dmcontent/govuk_frontend.py
```diff
@@ -11,7 +11,18 @@
...
+
+def govuk_input(
+    question: Question, data: Optional[dict] = None, errors: Optional[dict] = None
+) -> dict:
+    """Create govukInput macro parameters from a text question"""
+
+    params = _params(question, data, errors)
+    params["classes"] = "app-text-input--height-compatible"
+
+    return params
```
katstevens, line 74: Ahh this is really neat. I love it.

tests/test_govuk_frontend.py
```diff
@@ -21,48 +21,17 @@ def question(self):
...
-                'This will help you to refer to your requirements\n'
-                '</div>\n'
-                '100 characters maximum'
-            },
-            "value": "Find an individual specialist",
-        }
+        assert govuk_input(question, data) == snapshot
 
-    def test_with_errors(self, question):
+    def test_with_errors(self, question, snapshot):
```
katstevens, line 40: Ah does `syrupy` match the test name exactly to the snapshot name then?

dmcontent/govuk_frontend.py
```diff
@@ -0,0 +1,152 @@
...
+        # can be styled like a normal paragraph with the following Sass
+        #
+        #     .app-hint--text {
+        #       @extend %govuk-body--m
+        #     }
+        hint_html += Markup('<div class="app-hint--text">\n')
+        hint_html += escape(question.question_advice)
+        hint_html += Markup("\n</div>")
+        if question.get("hint"):
+            hint_html += "\n"
```
domoscargin, line 138: My only concern with this is that question_advice can get quite involved ("cultural fit criteria advice", for example, encompasses multiple headings, paragraphs and lists).

The service manual recommends [limiting hint text to about 3 lines max](https://www.gov.uk/service-manual/design/designing-good-questions#how-to-add-help-text).

There's potential for this to be offputting to screenreader users, since inputs are describedBy the hint text.

Can't immediately think of an easy way to place the advice text outside of the hint span that doesn't involve a template/macro.

tests/test_govuk_frontend.py
lfdebrux: Yes, it's super magical

dmcontent/govuk_frontend.py
```diff
@@ -0,0 +1,152 @@
...
+
+    `from_question` takes a Question and returns the name of the govuk-frontend
+    macro to call and the parameters to call it with. Calling the macro with
+    the parameters is left to the app developer.
+
+    A little bit of Jinja magic is required for this; assuming you have called
+    `from_question` already and have the macro name and parameters, you need a
+    table of macro names to macros in a Jinja template:
+
+        {% set govuk_frontend = {
```
domoscargin, line 45: Is there a potential namespace issue here or is this fine?

tests/test_govuk_frontend.py
lfdebrux: It all gets stored in the file [tests/__snapshots__/test_govuk_frontend.ambr](https://github.com/alphagov/digitalmarketplace-content-loader/pull/90/files#diff-4c0fcb4e0ec337b656861d12c91eda30) (ambr for amber, very cute).

I've seen other similar tools that use a pure Python file, but the advantage of this file format seems to be that it includes class information and deals with custom classes slightly better?

However the main reason I chose `syrupy` over `pytest-snapshot` or `snapshottest` is that it is being actively maintained, and it worked first time 😆

domoscargin:
I think this looks good (with comments) - and with this in place it feels like we should hopefully speed through the form tickets coming up.

dmcontent/govuk_frontend.py
lfdebrux: I thought about this, but I'm not sure I understand whether it makes a difference to a screenreader whether the question advice is in a separate div from the hint or not. As far as I can tell hint text is not a semantic thing, so the screenreader is not going to treat it specially from any other text on the page, right?

So if we're putting a huge block of text between the label and the input it's going to get read the same way regardless of the semantic layout, in theory? (I'm not sure about this, so testing with alphagov/digitalmarketplace-briefs-frontend#337 would be much appreciated).

All this is to say that the in practice the way to follow the service manual guidance is to get rid of the question advice, or severely cut it down... which isn't really a technical issue so outside the scope of this PR 😉

dmcontent/govuk_frontend.py
lfdebrux: 🤔 I don't think so... let's call it something else though just in case 😆

lfdebrux:
> I think this looks good (with comments) - and with this in place it feels like we should hopefully speed through the form tickets coming up!

That's the dream!🤞

dmcontent/govuk_frontend.py
lfdebrux: > ("cultural fit criteria advice", for example, encompasses multiple headings, paragraphs and lists).

😢

domoscargin: Inputs have an `aria-describedby` which gets set to the error and hint text if they're present. On Mac VoiceOver, this equates to inputs being read out as [question_label] "Edit Text" [Error text, Hint Text] "You are currently on...", so that's the concern.

However, as you say, that's not really in the scope of this PR, and I'd say this probably qualifies as accessible as is - we're certainly not excluding any information.

lfdebrux: > Inputs have an aria-describedby which gets set to the error and hint text if they're present. On Mac VoiceOver, this equates to inputs being read out as [question_label] "Edit Text" [Error text, Hint Text] "You are currently on...", so that's the concern.

Hmm, so does that mean if I'm tabbing through the page it will read the hint text twice?

It might be worth taking this to the accessibility team or the design system team. I would like to get this aspect right first time, as much as possible.

katstevens: I would treat questions like `culturalFitCriteria` as an exception, that we should be planning to remove/change the format of anyway. We know those lengthy questions are not a great experience for users, and ideally that huge amount of content would be elsewhere entirely (eg the guidance). So let's not make things overly complicated in order to cater for them.

domoscargin: Not an issue if you're tabbing through - you'd only hear it once on the input since you wouldn't tab onto the hint text itself.

It does repeat if you're letting the screenreader read the page for you.

Asked the Design System team:

> I don't think the design system itself has guidance on how long hint text should be, but we've definitely had conversations with people along similar lines where, for example, we've suggested that a long paragraph or bullet point list may be stretching the use of the hint text.
> The way the input component is set up, the input is describedby the hint text, so I imagine long hint text may be offputting for screenreader users

lfdebrux: > It does repeat if you're letting the screenreader read the page for you.

😬

lfdebrux: I've spiked having the label separate from the rest of the component, allowing us to drop the question advice in the middle. This seems to work, but the downside is some of the complexity gets pushed back into the apps.

https://github.com/alphagov/digitalmarketplace-briefs-frontend/compare/ldeb-govuk-input...ldeb-govuk-input--separate-label

Thoughts?

domoscargin: Of course, this brings up the opposite issue of users tabbing through and not having the advice for context 😫 

As long as the advice is semantic I think that's fine. I think I like this better? This way you won't have the potential for 20 repeats on something like the List Input component.

domoscargin: One issue: if the question advice is non-html, it's currently being inserted as a stray bit of text:

Screenshot 2020-08-11 at 15 56 46

lfdebrux: I'll look at how to make sure the question advice is always in `p` tag(s). You mentioned the question advice should be semantic, does that mean we should wrap it all in a `div` of some class?

domoscargin: Nope, as long as the advice follows standard hierarchy (<h2>, <p>, etc). If any of them get weirder than that we could wrap it, but I think they're all either plain text or html text content.

lfdebrux: > One issue: if the question advice is non-html, it's currently being inserted as a stray bit of text:

#92 should fix that

katstevens: This all sounds sensible to me. If the content markup in the frameworks is proving difficult to display, we have some flexibility to change it ourselves (changing the content itself is a bit more involved though).

lfdebrux: I've pushed the spike branch over this branch, so now the label is separate from the input component itself

lfdebrux:
I've pushed an update with a different design so we can handle question advice in a better way.

PR #92, opened 2020-08-11
Change Question.question_advice to be always Markdown

lfdebrux:
We're having issues where the question advice is sometimes wrapped in HTML `p` tags, and sometimes isn't.

Ideally it would be consistent, so this commit changes the behaviour of the content loader to always treat the `question_advice` field as Markdown.  This means it will always be HTML `<p>`s!

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '7.11.0'
+__version__ = '7.11.1'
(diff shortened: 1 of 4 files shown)


lfdebrux:
> 👍
> 
> The spacing's a bit large within the 'question-advice' wrapping span. Not sure how widespread this case is and how bothered we are with a temporary spacing issue, but might require CSS overrides per app.

lfdebrux:
> The spacing's a bit large within the 'question-advice' wrapping span. Not sure how widespread this case is and how bothered we are with a temporary spacing issue, but might require CSS overrides per app.

I think that has been present for a while now so I'm not too worried about it. As you say it is (hopefully) temporary, as we will be getting rid of the span tags :D

PR #93, opened 2020-08-14
Add List Input to dmcontent.govuk_frontend

domoscargin:
https://trello.com/c/VG4caYdJ/136-3-replace-list-form-with-list-input-component-in-create-a-dos-opportunity-journey

For Questions of type "list", we use the new List Input component.

Because this uses a fieldset legend as the question heading, we have less flexibility in where we can place it since it must wrap around its fields. 

Because we can alter List Input to suit our needs, we can simply account for question advice within the component itself. (See https://github.com/alphagov/digitalmarketplace-govuk-frontend/pull/163).

GOVUK components like Checkboxes will be trickier to handle, but I've left that as out of scope for this ticket.

This is for use on the Briefs FE: https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/339

### DIFF
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -70,6 +70,11 @@ def from_question(
             "macro_name": "govukInput",
             "params": govuk_input(question, data, errors),
         }
+    elif question.type == "list":
+        return {
+            "macro_name": "dmListInput",
+            "params": dm_list_input(question, data, errors)
+        }
     else:
         return None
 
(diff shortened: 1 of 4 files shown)


dmcontent/govuk_frontend.py
```diff
@@ -141,6 +178,19 @@ def _params(
...
+    # but checkboxes, radios and multiquestions will probably join
+    # this check.
+    if question.get("type") == "list":
+        params["fieldset"] = {
+            "legend": {
+                "text": get_label_text(question),
+                "isPageHeading": True,
+                "classes": "govuk-fieldset__legend--l"
+            }
+        }
```
lfdebrux, line 192: I see the advantage of having a separate function `get_label_text` for the text part of a legend.

Can I suggest we also have a separate function `govuk_fieldset`that generates the parameters for the fieldset, rather than adding it to `_params`?

We can add the params to the list input params easily with `params["fieldset"] = govuk_fieldset(question)`, and we can worry about what to do with other components later.

domoscargin: Done!

dmcontent/govuk_frontend.py
```diff
@@ -70,6 +70,11 @@ def from_question(
             "macro_name": "govukInput",
             "params": govuk_input(question, data, errors),
         }
+    if question.type == "list":
```
lfdebrux, line 73: This should be an `elif` (easy mistake to make! I'm trying to think how we could write a test to catch it 🤔 )

```suggestion
    elif question.type == "list":
```

lfdebrux: Actually I've just realised in this case it doesn't matter as we're returning None either way, but better to be safe than sorry!

PR #94, opened 2020-08-19
Add is_page_heading property

domoscargin:
For multiquestions, we don't want each child's label/legend to be set as the page heading, so we need a way to flag this.

This enables the is_page_heading property to be set as appropriate.

### DIFF
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -19,7 +19,7 @@ __all__ = ["from_question", "govuk_input", "govuk_label"]
 
 
 def from_question(
-    question: Question, data: Optional[dict] = None, errors: Optional[dict] = None
+    question: Question, data: Optional[dict] = None, errors: Optional[dict] = None, **kwargs
 ) -> Optional[dict]:
     """Create parameters object for govuk-frontend macros from a question
 
@@ -66,21 +66,21 @@ def from_question(
     """
     if question.type == "text":
         return {
-            "label": govuk_label(question),
+            "label": govuk_label(question, **kwargs),
             "macro_name": "govukInput",
-            "params": govuk_input(question, data, errors),
+            "params": govuk_input(question, data, errors, **kwargs),
         }
     elif question.type == "list":
         return {
             "macro_name": "dmListInput",
-            "params": dm_list_input(question, data, errors)
+            "params": dm_list_input(question, data, errors, **kwargs)
         }
     else:
         return None
 
 
 def govuk_input(
-    question: Question, data: Optional[dict] = None, errors: Optional[dict] = None
+    question: Question, data: Optional[dict] = None, errors: Optional[dict] = None, **kwargs
 ) -> dict:
     """Create govukInput macro parameters from a text question"""
 
@@ -91,7 +91,7 @@ def govuk_input(
 
 
 def dm_list_input(
-    question: Question, data: Optional[dict] = None, errors: Optional[dict] = None
+    question: Question, data: Optional[dict] = None, errors: Optional[dict] = None, **kwargs
 ) -> dict:
     """Create dmListInput macro parameters from a list question"""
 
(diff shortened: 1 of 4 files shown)


dmcontent/govuk_frontend.py
```diff
@@ -119,26 +119,31 @@ def dm_list_input(
     return params
 
 
-def govuk_label(question: Question) -> dict:
-
+def govuk_label(question: Question, is_page_heading: bool = True, **kwargs) -> dict:
```
lfdebrux, line 122: I think we should make `is_page_heading` kwargs only?

```suggestion
def govuk_label(question: Question, *, is_page_heading: bool = True, **kwargs) -> dict:
```

dmcontent/govuk_frontend.py
```diff
@@ -119,26 +119,31 @@ def dm_list_input(
...
+        # https://design-system.service.gov.uk/patterns/question-pages/
+        "isPageHeading": is_page_heading,
 
         "for": f"input-{question.id}",
         "text": get_label_text(question),
     }
 
 
-def govuk_fieldset(question: Question) -> dict:
+def govuk_fieldset(question: Question, is_page_heading: bool = True, **kwargs) -> dict:
```
lfdebrux, line 138: I think we should make `is_page_heading` kwargs only?

```suggestion
def govuk_fieldset(question: Question, *, is_page_heading: bool = True, **kwargs) -> dict:
```

dmcontent/govuk_frontend.py
```diff
@@ -119,26 +119,31 @@ def dm_list_input(
...
-
+def govuk_label(question: Question, is_page_heading: bool = True, **kwargs) -> dict:
+    """
+    :param bool is_page_heading: If True, the label will be set to display as a page heading
+    """
     return {
-        # Style the label as a page heading, following the
-        # GOV.UK Design System question pages pattern at
-        # https://design-system.service.gov.uk/patterns/question-pages/
         "classes": "govuk-label--l",
```
lfdebrux, line 127: Do we want this style if it's not a page heading?

dmcontent/govuk_frontend.py
```diff
@@ -119,26 +119,31 @@ def dm_list_input(
...
+    """
+    :param bool is_page_heading: If True, the legend will be set to display as a page heading
+    """
 
     return {
         "legend": {
             "text": get_label_text(question),
-            "isPageHeading": True,
+            "isPageHeading": is_page_heading,
             "classes": "govuk-fieldset__legend--l"
```
lfdebrux, line 147: Do we want this style if it's not a page heading?

dmcontent/govuk_frontend.py
domoscargin: Ooh, I don't actually know what this means. I'm guessing the asterisk is placed to handle any non-keyworded args?

So if you were to pass in, say, (question, is_page_heading=False, "A random string"), would that fail or be handled?

dmcontent/govuk_frontend.py
domoscargin: Good catch!

dmcontent/govuk_frontend.py
lfdebrux: The syntax is a little obtuse, but the meaning is that you can't accidentally set `is_page_heading` by passing in two arguments. Without the asterisk, `govuk_fieldset(question, question)` would be valid. Granted it's unlikely to happen or be a problem, but better safe than sorry! I find it also makes things clearer when people write the name of the keyword argument (rather than `govuk_fieldset(question, True)`. What is true doing there?) and this change enforces that.

This feature was added in [Python Enhancement Proposal 3102](https://www.python.org/dev/peps/pep-3102/), but this article explains it a little better https://python-3-for-scientists.readthedocs.io/en/latest/python3_advanced.html#keyword-only-arguments.

lfdebrux: Incidentally the example you gave would fail either way.

lfdebrux: So in answer to your question, the `*` syntax prevents any further positional arguments from being read after that point in the argument list.

lfdebrux (approved):

PR #97, opened 2020-08-20
Add support for radios questions to dmcontent.govuk_frontend

lfdebrux:
Ticket: https://trello.com/c/6xOXfwx7/123-2-use-radios-component-for-radios-questions-in-create-a-dos-opportunity-journey

This adds `govuk_radios` to create params for govukRadio from Question objects, and updates `from_question` to handle that question type.

This uses the [function `govuk_options` from dmutils](https://github.com/alphagov/digitalmarketplace-utils/blob/79cbc5eac674cb438fc90c598b76395f50491a25/dmutils/forms/helpers.py#L70) to do most of the work (thanks @agalamatis!).

It doesn't support questions with type boolean or items with conditionally revealing content, but as we don't use those in the create a DOS opportunity journey I considered that out of scope.

### DIFF
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -11,6 +11,7 @@ Read the docstring for `from_question` for more detail on how this works.
 from typing import Optional
 
 from dmutils.forms.errors import govuk_error
+from dmutils.forms.helpers import govuk_options
 
 from dmcontent.questions import Question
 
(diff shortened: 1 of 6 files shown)


dmcontent/govuk_frontend.py
```diff
@@ -75,6 +76,12 @@ def from_question(
             "macro_name": "dmListInput",
             "params": dm_list_input(question, data, errors, **kwargs)
         }
+    elif question.type == "radios":
+        return {
+            "fieldset": govuk_fieldset(question, **kwargs),
```
domoscargin, line 81: 🙌

domoscargin (approved):
Looking good and working for me locally!

PR #98, opened 2020-08-21
Add support for checkboxes questions to dmcontent.govuk_frontend

domoscargin:
https://trello.com/c/zx1hJsgX/114-2-replace-checkboxes-in-create-a-dos-opportunity-journey

Builds on the work done in #97, adding `govuk_checkboxes` to create params for govukCheckboxes from Question objects, and updates `from_question` to handle that question type.

### DIFF
--- a/tests/test_govuk_frontend.py
+++ b/tests/test_govuk_frontend.py
@@ -6,6 +6,7 @@ from dmcontent.govuk_frontend import (
     from_question,
     govuk_character_count,
     govuk_input,
+    govuk_checkboxes,
     govuk_radios,
     dm_list_input,
     govuk_fieldset,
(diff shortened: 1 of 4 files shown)


tests/test_govuk_frontend.py
```diff
@@ -138,6 +139,81 @@ def test_from_question_with_errors(self, question, snapshot):
...
 
+class TestCheckboxes:
+    @pytest.fixture
+    def question(self):
+        return Question(
+            {
+                "id": "yesOrNo",
+                "name": "Yes or no",
+                "question": "Yes or no?",
+                "type": "radios",
```
katstevens, line 150: Will this be `type: checkboxes` in practice? (I don't think it matters for this test)

katstevens (approved):
One small comment, don't think it's blocking

tests/test_govuk_frontend.py
domoscargin: Oof. Thanks for catching!

PR #99, opened 2020-08-24
Add support for date questions to dmcontent.govuk_frontend

domoscargin:
https://trello.com/c/k2SADNQr/172-2-replace-date-input-component-in-create-a-dos-opportunity-journey

Adds govuk_date_input to create params for govukDateInput from Question objects, and updates from_question to handle that question type.

## Notes
We don't currently do granular error checking for date inputs, and this component reflects that. The [Design System component](https://design-system.service.gov.uk/components/date-input/) does recommend error messages are specific, but I think that's out of scope for this ticket.

Bumped version to 7.18 because the checkboxes PR (#98) is ready to go and I'll merge that first as 7.17.

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '7.17.0'
+__version__ = '7.18.0'
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -71,6 +71,12 @@ def from_question(
             "macro_name": "govukInput",
             "params": govuk_input(question, data, errors, **kwargs),
         }
+    if question.type == "date":
+        return {
+            "fieldset": govuk_fieldset(question, **kwargs),
+            "macro_name": "govukDateInput",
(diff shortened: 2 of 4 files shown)


dmcontent/govuk_frontend.py
```diff
@@ -103,6 +109,41 @@ def govuk_input(
     return params
 
 
+def govuk_date_input(
+    question: Question, data: Optional[dict] = None, errors: Optional[dict] = None, **kwargs
+) -> dict:
+    """Create govukRadios macro parameters from a radios question"""
```
katstevens, line 115: Docstring will need updating...

dmcontent/__init__.py
```diff
@@ -2,4 +2,4 @@
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '7.16.0'
+__version__ = '7.18.0'
```
katstevens, line 5: Did we skip `7.17`?

katstevens (approved):
Looks good - two small comments

dmcontent/__init__.py
domoscargin: I've pre-empted the checkboxes PR going in first as 7.17

katstevens: 😹 I thought I'd approved that one already - will do so now!

PR #102, opened 2020-09-02
Add support for number questions to govuk_frontend.from_question

lfdebrux:
Ticket: https://trello.com/c/0cIr1131/210-2-replace-number-type-questions-in-create-a-dos-opportunity-journey

We can now use GOV.UK Frontend text input components for number questions.

#### Before

Screenshot 2020-09-02 at 15 47 47

#### After

Screenshot 2020-09-02 at 15 46 55

### DIFF
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -65,7 +65,7 @@ def from_question(
     :returns: A dict with the macro name, macro parameters, and labels, or None
               if we don't know how to handle this type of question
     """
-    if question.type == "text":
+    if question.type == "text" or question.type == "number":
         return {
             "label": govuk_label(question, **kwargs),
             "macro_name": "govukInput",
@@ -107,10 +107,27 @@ def from_question(
 def govuk_input(
     question: Question, data: Optional[dict] = None, errors: Optional[dict] = None, **kwargs
 ) -> dict:
-    """Create govukInput macro parameters from a text question"""
+    """Create govukInput macro parameters from a text or number question"""
 
     params = _params(question, data, errors)
     params["classes"] = "app-text-input--height-compatible"
+
+    if question.type == "number":
+        params["classes"] += " govuk-input--width-5"
+        params["spellcheck"] = False
+        if question.get("limits") and question.limits.get("integer_only") is True:
+            params["inputmode"] = "numeric"
+            params["pattern"] = "[0-9]*"
+
+        if question.get("unit"):
+            prefix_or_suffix = {"before": "prefix", "after": "suffix"}[question.unit_position]
+            params[prefix_or_suffix] = {"text": question.unit}
+            if params.get("pattern"):
+                unit_regex = f"{question.unit}?"
+                if prefix_or_suffix == "prefix":
+                    params["pattern"] = unit_regex + params["pattern"]
+                elif prefix_or_suffix == "suffix":
+                    params["pattern"] += unit_regex
 
     return params
 
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -264,6 +264,8 @@ def govuk_character_count(
 ) -> dict:
     params = _params(question, data, errors)
 
+    params["spellcheck"] = True
+
     if question.get("max_length_in_words"):
         params["maxwords"] = question.max_length_in_words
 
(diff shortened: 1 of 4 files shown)


lfdebrux:
This PR includes the prefix and suffix params, but they won't have any effect until alphagov/govuk-frontend#1915 is merged and the app is using the latest version of govuk-frontend v3.

Using govuk-frontend v2 the text inputs will not show any prefix or suffix, for example:

Screenshot 2020-09-02 at 15 51 45

dmcontent/govuk_frontend.py
```diff
@@ -107,11 +107,30 @@ def from_question(
...
     question: Question, data: Optional[dict] = None, errors: Optional[dict] = None, **kwargs
 ) -> dict:
-    """Create govukInput macro parameters from a text question"""
+    """Create govukInput macro parameters from a text or number question"""
 
     params = _params(question, data, errors)
     params["classes"] = "app-text-input--height-compatible"
 
+    if question.type == "number":
+        params["classes"] += " govuk-input--width-5"
```
domoscargin, line 116: I'm guessing this is all that's used on Briefs? (ie: we don't have to allow for different widths right now unless they pop up on other apps?).

dmcontent/govuk_frontend.py
```diff
@@ -107,11 +107,30 @@ def from_question(
...
+        params["spellcheck"] = False
```
domoscargin, line 117: 👍 

Don't think there's many instances of us using text inputs for anything which would need a spellcheck.

Perhaps we should add it to the textareas though...

dmcontent/govuk_frontend.py
```diff
@@ -107,11 +107,30 @@ def from_question(
...
+        if question.get("limits") and question.limits.get("integer_only") is True:
+            params["inputmode"] = "numeric"
+            params["pattern"] = "[0-9]*"
+        else:
+            params["inputmode"] = "text"
```
domoscargin, line 122: Do we ever ask for decimal numbers?

If so, we'd have to not include `inputmode` and `pattern` as per https://design-system.service.gov.uk/components/text-input/#asking-for-decimal-numbers

dmcontent/govuk_frontend.py
```diff
@@ -107,11 +107,30 @@ def from_question(
...
+
+        if question.get("unit"):
+            prefix_or_suffix = {"before": "prefix", "after": "suffix"}[question.unit_position]
```
domoscargin, line 125: 🙌

tests/test_govuk_frontend.py
```diff
@@ -65,6 +65,122 @@ def test_with_errors(self, question, snapshot):
...
+
+        https://design-system.service.gov.uk/components/text-input/#numbers
+        """
+
+        if integer_only is False:
+            question.limits = {"integer_only": False}
+
+        params = govuk_input(question)
+
+        assert params["inputmode"] == "text"
```
domoscargin, line 106: Decimal numbers should not include `inputmode`

https://design-system.service.gov.uk/components/text-input/#asking-for-decimal-numbers

dmcontent/govuk_frontend.py
lfdebrux: Yeah, the only places we use it the size is currently just `smaller` 😆 https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/forms/macros/forms.html#L232

dmcontent/govuk_frontend.py
lfdebrux: I just copied from the Design System examples haha https://design-system.service.gov.uk/components/text-input/#numbers

Good idea adding `spellcheck=True` to textareas though!

dmcontent/govuk_frontend.py
lfdebrux: Yeah, decimal numbers is the `else` branch. In the schema for our question YAML files it is assumed to be decimal unless `integer_only` is true-ish https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/schema_generator/validation.py#L374

tests/test_govuk_frontend.py
lfdebrux: Ah right I misread that! Good spot, I'll fix that.

lfdebrux: Fixed

dmcontent/govuk_frontend.py
lfdebrux: I've added that to the `govuk_character_count` function

domoscargin (approved):
🔢 

Looks good!

PR #103, opened 2020-09-04
Add height compatible override class to date input inputs

lfdebrux:
We want to make sure the inputs are all the same height in our apps.

#### After

image


domoscargin (approved):
Tests are good, working locally. Good spot!

PR #104, opened 2020-09-04
Add dm_pricing_input to govuk_frontend

lfdebrux:
https://trello.com/c/oIEbMdvM/213-2-replace-pricing-type-questions-in-create-a-dos-opportunity-journey

This only handles the case where the pricing question has a single field `price`, as this is the only case found in the briefs frontend. However I've taken some care to make sure the design will extend to the multiple fields case.

Like with the number field, this commit includes the params for a prefix, but they won't show up until https://github.com/alphagov/govuk-frontend/issues/1915 is closed.

#### Before

Screenshot 2020-09-04 at 15 23 57

### DIFF
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -107,12 +107,12 @@ def from_question(
 def govuk_input(
     question: Question, data: Optional[dict] = None, errors: Optional[dict] = None, **kwargs
 ) -> dict:
-    """Create govukInput macro parameters from a text or number question"""
+    """Create govukInput macro parameters from a text, number or pricing question"""
 
     kwargs.setdefault("classes", ["app-text-input--height-compatible"])
     params = _params(question, data, errors, **kwargs)
 
-    if question.type == "number":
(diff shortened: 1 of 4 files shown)


dmcontent/govuk_frontend.py
```diff
@@ -224,19 +233,55 @@ def dm_list_input(
...
+                data,
+                errors,
+                input_id=question.fields["price"],
+                prefix_text="£",
+            ),
+        }
+
+    # TODO: Handle pricing questions with multiple fields. For now (for the
+    # briefs frontend) we only need to handle the simple case. Have a look at
+    # branch ldeb-spike-pricing-input-multiple-fields for a sketch of how to do
```
domoscargin, line 262: I can't seem to find this branch, but this looks good to me.

domoscargin (approved):
LGTM

dmcontent/govuk_frontend.py
lfdebrux: Ah, thanks for the reminder, I've pushed the branch now

PR #105, opened 2020-09-07
Add feature flag to Question to change behaviour of href

lfdebrux:
If an app is using govuk-frontend version 3 and up we don't want the href to end with a `-1`, as that is not what the template will output.

To use this an app should set `dmcontent.govuk_frontend.govuk_frontend_version` at some point early on in the loading process.


domoscargin (approved):
Looks good!

PR #106, opened 2020-09-07
Update dependencies

lfdebrux:
Some dependency wrangling:

- Update version of digitalmarketplace-content-loader we test against to match version required by `setup.py`
- Add requirements.txt frozen by `pip-compile` so we're always testing with the same versions of dependencies
- Allow using this library with Jinja 2.11


domoscargin (approved):

katstevens:
Does having a `requirements.in` mean we can add this repo to dependabot? Should we be doing it for our other libraries?

lfdebrux:
> Does having a `requirements.in` mean we can add this repo to dependabot?

We could and we should!

Although I think it will only update the requirements.txt file, not the setup.py file, so I don't know if it will help with the [RFC on managing dependencies for our Python libraries](https://docs.google.com/document/d/1i3Pz29C1M1C0G51q6P2QWSy85oolneOcFiEdwDo74oo/edit#heading=h.ewttup9cajgj). I will check it out though.

> Should we be doing it for our other libraries?

We should probably do this for other libraries though, in this case I realised that not doing it meant that I couldn't be sure what version of Jinja was being tested, which was bad because I wanted to be sure it would work with Jinja 2.11. Adding the `requirements.in` file and freezing to `requirements.txt` helped make it explicit and fixed, and I think that it would be similar for other repos. I can make the change.

PR #107, opened 2020-09-09
Add GitHub Dependabot configuration

lfdebrux:
I'm trying an experiment into managing library dependencies using Dependabot, for the open RFC https://docs.google.com/document/d/1i3Pz29C1M1C0G51q6P2QWSy85oolneOcFiEdwDo74oo/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+# Dependabot config
+# https://dependabot.com/docs/config-file/
+# Checks and updates dependencies in
+# `requirements.in`, `requirements.txt`,
+# and `setup.py`
+
+version: 2
+updates:
+  - package-ecosystem: pip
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
```
katstevens, line 1: Can we include a comment at the top saying which files this should be checking (similar to [the other repos](https://github.com/alphagov/digitalmarketplace-scripts/blob/master/.dependabot/config.yml))?

lfdebrux: Ah yeah, good spot

lfdebrux: Done

katstevens (approved):
Let's try it 👍

PR #108, opened 2020-09-23
Add support for pricing without minima

bjgill:
Trello: https://trello.com/c/yWkgqyPL/325-2-update-dos5-application-content-in-the-frameworks-repo

In DOS 5, we will only have the maximum. Update the summary so this gets handled correctly. Add tests to verify. Add in some type hints for fun.

I am assuming that nothing that currently exists relies on the previous behaviour. As far as I can tell, form verification ensures that the minimum is provided where it is needed.

### DIFF
--- a/dmcontent/formats.py
+++ b/dmcontent/formats.py
@@ -1,16 +1,31 @@
 # -*- coding: utf-8 -*-
+from typing import Optional, Union
 
 
-def format_price(min_price, max_price, unit, interval, hours_for_price=None):
+def format_price(min_price: Optional[Union[str, float]],
+                 max_price: Optional[Union[str, float]],
+                 unit: Optional[str],
+                 interval: Optional[str],
+                 hours_for_price: Optional[str] = None):
     """Format a price string"""
-    if hours_for_price:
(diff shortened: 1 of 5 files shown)


dmcontent/formats.py
```diff
@@ -1,15 +1,25 @@
...
 
-def format_price(min_price, max_price, unit, interval, hours_for_price=None):
+def format_price(min_price: Optional[Union[str, float]],
+                 max_price: Optional[Union[str, float]],
+                 unit: Optional[str],
+                 interval: Optional[str],
+                 hours_for_price: Optional[str] = None):
     """Format a price string"""
+    if min_price is not None:
+        base_price = min_price
```
lfdebrux, line 12: This could allow the min_price to be zero, or the empty string, right?

bjgill: Yes. The empty string is obviously nonsense, though. I've just pushed an updated version that rejects the empty string. I think it also makes the logic more obvious.

lfdebrux (approved):
Neat! 🔥

PR #110, opened 2020-10-06
Start testing against Python 3.9

bjgill:
And fix a DeprecationWarning: "Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ python:
   - "3.6"
   - "3.7"
   - "3.8"
+  - "3.9"
 install:
   - make requirements-dev
 script:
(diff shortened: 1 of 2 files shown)


bjgill:
Ah. 3.9 isn't yet fully released on travis

gidsg:
we could try switching to github actions? 😉

lfdebrux:
Look at all those Python versions we support! 😁

lfdebrux (approved):

bjgill:
Indeed. I'd like to start dropping support for the older versions, but we probably don't have time to schedule that work just yet, alas.

PR #111, opened 2020-10-07
Update Markdown requirement to latest major version

lfdebrux:
We want to be able to keep up to date with [Python-Markdown](https://pypi.org/project/Markdown/), which is currently on version 3.3. Currently however we are constrained to use the latest 2 series version, which was released in 2018.

This PR adds tests to our Markdown extension so we can be sure nothing will break from upgrading, and lifts the upper bound on the requirement constraint.

This PR does not change the extension (although it uses some deprecated APIs) because that would be a breaking change, which we don't want right now.

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '7.23.0'
+__version__ = '7.24.0'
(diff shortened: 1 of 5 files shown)


bjgill (approved):
Looks sensible. I assume none of the [breaking changes to markdown](https://python-markdown.github.io/change_log/release-3.0/) apply to us?

lfdebrux:
> Looks sensible. I assume none of the [breaking changes to markdown](https://python-markdown.github.io/change_log/release-3.0/) apply to us?

No, we don't use any of the affected features.

PR #112, opened 2020-10-07
Add remove prefix/suffix support to number type questions

lfdebrux:
Ticket: https://trello.com/c/MwEMI2pR/428-fix-the-problem

We want to be able to gracefully handle the case where users input units into a form, following Design System guidance [[1]].

Now number type questions with a unit will be automatically filter those units from the user data.

The code to remove prefixes or suffixes is copied from [PEP-616][2], one day we might be able to replace it with the `removeprefix()` and `removesuffix()` methods.

[1]:
https://design-system.service.gov.uk/components/text-input/#prefixes-and-suffixes
[2]: https://www.python.org/dev/peps/pep-0616/

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '7.24.0'
+__version__ = '7.25.0'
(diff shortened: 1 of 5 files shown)


bjgill (approved):
Neat. This looks like a really nice way of solving the problem.

I assume we could extend this in future to handle questions with `type: pricing`?

lfdebrux:
> Neat. This looks like a really nice way of solving the problem.
> 
> I assume we could extend this in future to handle questions with `type: pricing`?

Yep, I think so.

PR #115, opened 2020-10-20
Add commas to price formatting for G-Cloud services

alex9smith:
Add commas to the prices for G-cloud services so they're easier to read, and bump the minor version number

Addresses https://trello.com/c/UL5WzBD7/1412-add-commas-to-price-formatting-on-g-cloud-service-pages

### DIFF
--- a/dmcontent/formats.py
+++ b/dmcontent/formats.py
@@ -13,13 +13,13 @@ def comma_format(number: Union[str, float, int]) -> str:
     :param number: The number to be formatted. Either an int, a float or a str.
     :return: A string of the comma formatted number
     """
-    if type(number) == str:
+    if isinstance(number, str):
         try:
             number = int(number)
         except ValueError:
             number = float(number)
 
-    if type(number) == int:
+    if isinstance(number, int):
         # Don't need to add decimal places to an int
         return f"{number:,}"
     else:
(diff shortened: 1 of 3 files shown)


dmcontent/formats.py
```diff
@@ -2,16 +2,40 @@
...
+    Format a number with commas if it's greater than 999.
+    eg: `1000 -> '1,000'
+        123456.78 -> '123,456.78'
+        '1000000' -> '1,000,000'
+        500 -> '500'
+
+    :param number: The number to be formatted. Either an int, a float or a str.
+    :return: A string of the comma formatted number
+    """
+    if type(number) == str:
```
bjgill, line 16: Is it necessary to provide exactly `str`? Could we use [`isinstance`] to allow for subclasses of `str`?

(and ditto for `int` below)

bjgill (commented):
Looks sensible - I particularly like the test.

Just one question about the type testing and whether we can be more permissive.

dmcontent/formats.py
alex9smith: That does seem sensible. I'll have a look at where this gets called to check we're not doing anything unusual with types there but if not I'll update

bjgill (approved):
:shipit:

PR #118, opened 2020-11-19
Choose whether to use 'a' or 'an'

bjgill:
Trello: https://trello.com/c/RlNy5WlG/2074-ccsrequests-re-listing-correction

Previously, we always used 'a'. This doesn't work for things like 'a instance', and a supplier has noticed this. Fixing this properly will be a bit involved. So add a heuristic that should work most of the time.

If this is not enough, we should just use inflect rather than trying to write code that understands english better than this.

This is hacky as all heck, so I've included a comment to try and explain this monstrosity to future devs.


gidsg (approved):
👍 🅰️

PR #119, opened 2020-12-03
Add boolean to govuk_frontend

domoscargin:
https://trello.com/c/13tOvxuu/285-2-replace-content-loader-form-macros-in-brief-responses-frontend

digitalmarketplace-frontend-toolkit uses `boolean` questions, which are yes/no inline radio inputs:

Screenshot 2020-12-03 at 17 44 48

We don't currently handle them with govuk_frontend, but our work on `brief-responses` has uncovered some:
`yesNo`
`essentialRequirementsMet`

We can simply use `govukRadios` and add `govuk-radios--inline` to the `classes` property to achieve the effect in govuk-frontend.

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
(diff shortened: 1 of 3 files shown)


lfdebrux (approved):
Great stuff

PR #121, opened 2020-12-07
Add options to boolean question

domoscargin:
https://trello.com/c/13tOvxuu/285-2-replace-content-loader-form-macros-in-brief-responses-frontend

During testing on Brief Responses, I discovered that boolean questions don't have an `options` property, and so fail.

This fixes that by setting the options property.

### DIFF
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -215,8 +215,10 @@ def govuk_radios(
             params["classes"] += " govuk-radios--inline"
         else:
             params["classes"] = "govuk-radios--inline"
-
-    params["items"] = govuk_options(question.options, data.get(question.id))
+        options = [{"label": "Yes", "value": "True"}, {"label": "No", "value": "False"}]
+        params["items"] = govuk_options(options, str(data.get(question.id)))
+    else:
+        params["items"] = govuk_options(question.options, data.get(question.id))
 
     return params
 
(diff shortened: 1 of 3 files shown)


dmcontent/govuk_frontend.py
```diff
@@ -215,8 +215,10 @@ def govuk_radios(
             params["classes"] += " govuk-radios--inline"
         else:
             params["classes"] = "govuk-radios--inline"
-
-    params["items"] = govuk_options(question.options, data.get(question.id))
+        options = [{"label": "Yes", "value": "True"}, {"label": "No", "value": "False"}]
+        params["items"] = govuk_options(options, f"{data.get(question.id)}")
```
lfdebrux, line 219:
```suggestion
        params["items"] = govuk_options(options, data.get(question.id))
```

lfdebrux (approved):

dmcontent/govuk_frontend.py
domoscargin: Tried this initially, but it then fails in dm_utils - `govuk_options` is expecting this to be a string or dict of strings - `essentialRequirementsMet` on Brief Responses data property is:

`{'essentialRequirementsMet': True}`

If better to adjust the utils method, I can do that?

lfdebrux: Oh in that case I would suggest doing this: 

```suggestion
        params["items"] = govuk_options(options, str(data.get(question.id)))
```

It amounts to the same thing but is more explicit...

Or even more explicit would be 

```suggestion
        params["items"] = govuk_options(options, "True" if data.get(question.id) else "False")
```

but I'm fairly confident the string representation of true and false isn't changing any time soon....

PR #124, opened 2021-01-07
Allow date questions to unformat when there is no data

alex9smith:
Is DOS5 the first question when submitting a brief response is a date question, so the brief response has no data at that point. 

Also move unformat test with no data to base class so we know all subclasses of `Question` can handle the case with no data.


bjgill (approved):
:tada:

PR #125, opened 2021-01-08
Add `govuk_frontend.render_question()` for calling govuk-frontend macros from Python

lfdebrux:
Ticket: https://trello.com/c/fu3k1QO7/290-3-replace-content-loader-questions-with-followup-in-brief-responses-frontend

`render()` is a Python replacement for the [`show_question.html`][1] template that has developed to handle the output of `from_question()`.  While `show_question.html` is working for most cases, it has a couple of downsides

1. we have it repeated across all frontend repos where it's needed
2. it can't handle complex forms such as radios with conditionals
3. it's code is a little bit repetitive

This commit aims to address those problems by moving the logic to show a question into `dmcontent` where it can be shared across all apps, and by rewriting it in Python that should also make it possible to do more complicated things in future.

Currently `render()` is a bit more complicated than it needs to be, so that the output of `from_question()` can stay the same, but in future we can push more of the layout stuff into `from_question()` and make `render()` simpler and more general.

Apps should aim to use the convenience function `render_question()` which calls `from_question()` and `render()` in tandem, so that they don't need to worry about the interface between the two. Add it to your Jinja2 environment globals!

[1]: https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/blob/master/app/templates/macros/show_question.html

---

https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/260 shows this in action.

### DIFF
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -451,7 +451,7 @@ def _params(
 # `from_question()` we can safely change the objects emitted by `from_question()`
 # and experiment a bit more.
 
-@jinja2.contextfunction  # noqa: C901
+@jinja2.contextfunction
 def render(ctx, obj, *, question=None) -> Markup:
     """Call Jinja2 macros using Python objects
 
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -432,6 +432,10 @@ def render(ctx, obj, *, question=None) -> Markup:
     come up with a method of creating deferred call objects and a new function
     `render()` that evaluates the deferred calls.
 
+    `render()` still needs to be called from within a Jinja2 template, but it
+    should be able to handle expressions of arbitrary complexity (i.e. you can
+    have a macro call which calls other macros in its arguments).
+
     For convenience `render()` will also accept strings (either `str` or
     `Markup`) and pass them through, and if given a list of objects it will
     call itself on each object and join the result into one piece of HTML.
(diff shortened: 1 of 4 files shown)


dmcontent/govuk_frontend.py
```diff
@@ -1,15 +1,55 @@
...
+`render_question()` needs to be called from within a Jinja2 template, so it
+should be added to your template environment. You can add it to all your
+templates by adding it to your environment globals like so:
+
+    >>> import jinja2
+    >>> from dmcontent.govuk_frontend import render_question
+    >>> env = jinja2.Environment()
+    >>> env.globals["render_question"] = render_question
+
+You will also need to make sure all the GOV.UK (and Digital Marketplace) macros
```
domoscargin, line 26: Are these all required or could you just import, for example, govukInput if that's all you needed?

lfdebrux: You don't have to import them all, but you're in danger of getting a runtime error if you're not sure of what you're doing.

lfdebrux: Although now I think about it that's always true in Python/Jinja, calling a function that doesn't exist is a runtime error 😁

domoscargin (commented):
I think this all makes sense to me and is a neat way of avoiding the issues with using Jinja Macros across apps.

domoscargin (approved):
👍

tests/test_govuk_frontend.py
```diff
@@ -849,3 +856,156 @@ def test_value_and_error_message(self, question):
...
+                    question, data, errors, some_kwarg="foobar"
+                )
+                assert render.call_args == mock.call(
+                    ctx,
+                    from_question(question, data, errors, some_kwarg="foobar"),
+                    question=question
+                )
+
+    def test_it_renders_question_advice(self):
+        pass
```
alex9smith, line 1011: What's this for?

lfdebrux: 😱 I forgot to write that! Thanks! I wanted to test that it recreates [question advice functionality from `show_question.html`](https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/blob/master/app/templates/macros/show_question.html#L27) correctly

alex9smith: I figured it was either that, or there was some inheritance magic elsewhere that I was missing 😆

alex9smith (approved):
Looks good to me other than that empty test 👍

tests/test_govuk_frontend.py
lfdebrux: Gave the test some body

PR #127, opened 2021-01-13
Fix behaviour of render_question with unrecognised question type

lfdebrux:
Fixes a bug I found while trying to bump dmcontent in alphagov/digitalmarketplace-briefs-frontend#379.


gidsg (approved):
LGTM 👍

PR #132, opened 2021-02-01
Add support for followup questions in multiquestions

lfdebrux:
Ticket: https://trello.com/c/cMtAeSL2/764-3-add-support-for-conditionally-revealed-content-to-dmcontentgovukfronted

We want to be able to handle multiquestions that contain questoins with followups, like DOS5's [niceToHaveRequirements and yesNo questions][1]. The Design System calls this [conditionally revealing content][2].

This PR adds logic to `dm_multiquestion()` to handle these questions.

We put the logic here rather than in say `govukRadios()` because a) the way followups currently work is a little confusing, and b) it would be better in the long term to re-evaluate how we lay out these questions for the user and define them in the framework manifests. For now this commit gives us some form of backwards compatibility to unblock migration to govuk-frontend.

### Visual changes

This does change how these questions will look for users, in a few small ways:

- 'yes' or 'no' selection question will not have inline radios
- followup appears between 'yes' and 'no'
- labels for followup questions are not bold
- the space for followup questions is slightly smaller (this might just be a difference between frontend toolkit textbox and govuk-frontend character count components)

I think these differences are small enough that we shouldn't work around them however.

<table>
<thead>
<td>Before:</td>
<td>After:</td>
</thead>
<trow>
<td>
Screenshot 2021-02-01 at 15 02 56
</td>
<td>
Screenshot 2021-02-01 at 15 03 14
</td>
</trow>
</table>

[1]: https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists-5/questions/brief-responses/niceToHaveRequirements.yml
[2]: https://design-system.service.gov.uk/components/radios/#conditionally-revealing-content

### DIFF
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -315,10 +315,29 @@ def dm_multiquestion(
     if question.get("question_advice"):
         to_render.append(_question_advice(question))
 
-    to_render += [
-        from_question(q, data, errors, is_page_heading=False)
-        for q in question.questions
-    ]
+    to_skip = set()  # we need to be able to skip some questions, such as followups
+    for q in question.questions:
+        if q.id in to_skip:
+            continue
+
+        to_render.append(
+            from_question(q, data, errors, is_page_heading=False)
+        )
+
+        if q.get("followup"):
+            to_skip |= q.followup.keys()
+
+            # convert the values in values_followup to str
+            # to match the form input item values
+            followups = {str(v): qs for v, qs in q.values_followup.items()}
+            items = to_render[-1]["params"]["items"]
+
+            for item in items:
+                if item["value"] in followups:
+                    followup_q = question.get_question(followups[item["value"]][0])
+                    item["conditional"] = {
+                        "html": from_question(followup_q, data, errors, is_page_heading=False)
+                    }
 
     return to_render
 
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -45,7 +45,7 @@ Read the docstring for `from_question()` for more detail on how Questions are
 handled.
 """
 
-from typing import List, Optional, TYPE_CHECKING
+from typing import List, Optional, Set, TYPE_CHECKING
 
 import jinja2
 from jinja2 import Markup, escape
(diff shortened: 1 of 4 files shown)


dmcontent/govuk_frontend.py
```diff
@@ -315,10 +315,29 @@ def dm_multiquestion(
     if question.get("question_advice"):
         to_render.append(_question_advice(question))
 
-    to_render += [
-        from_question(q, data, errors, is_page_heading=False)
-        for q in question.questions
-    ]
+    to_skip = set()  # we need to be able to skip some questions, such as followups
+    for q in question.questions:
+        if q.id in to_skip:
```
gidsg, line 320: Sorry if I'm misunderstanding the control flow but given we instantiate this as an empty set above, how will this ever be true?

lfdebrux: Later on in the for loop we add things to the set using `

lfdebrux: If the syntax isn't clear I can change it to `to_skip.update(...)`

gidsg: I'm still slightly confused about the logic, since this happens first, sorry of I'm being slow, is this it stop it trying to add things more than once?

To be honest I prefer `.update` as symbol based operators are generally more difficult to google.

lfdebrux: No worries, it's good that you're questioning it, it needs to be clear.

Let me expand the comment.

lfdebrux: Is that any clearer?

tests/test_govuk_frontend.py
```diff
@@ -745,6 +798,42 @@ def test_from_question_with_errors(self, question, snapshot):
...
+                "href": "#input-answer",
+                "question": "Tell us more",
+                "message": "Enter an answer.",
+            },
+        }
+
+        form = from_question(question_with_followup, errors=errors)
+
+        assert "errorMessage" not in form[1]["params"]
+        assert form[1]["params"]["items"][0]["conditional"]["html"]["params"]["errorMessage"]
```
gidsg, line 833: Should we assert the expected error message text?

lfdebrux: Eh, I don't think it matters too much what the error message text is, just that it's there. In any case the snapshot assertion below will catch any situations where the message text is mangled; a representation of the value gets compared to the saved snapshot in https://github.com/alphagov/digitalmarketplace-content-loader/pull/132/files#diff-e9a82f20271c1113cecfad9c08777a4a380a669f4ed9a5b71c840323bec42125R581.

gidsg (approved):
👍 🔘

PR #134, opened 2021-02-19
Add type hints to `get_error_messages`

bjgill:
https://trello.com/c/9RUq5TL8/1243-2-bug-emessage-passed-to-inappropriate-error-handler

We have previously introduced bugs by passing the wrong type in. Add type hints to make this mistake harder to make in future.


lfdebrux (approved):

bjgill:
Merging. I don't think it's worth my time propagating this to the apps that use it.

PR #137, opened 2021-03-04
Allow lazy loading of manifests 

bjgill:
Manifests are slow to load - they take ~1 second each. This contributes to very slow app startup times - ~20-30 seconds for supplier-frontend.

Many of the manifests are for old frameworks that are requested very rarely or not at all. Thus, we can lazily load these old frameworks to give us a speedup on app startup without significantly slowing down real users.

See alphagov/digitalmarketplace-supplier-frontend#1348 for the experimental version of this I used for testing. Based on the numbers in that PR, I think this will save developers a few minutes per day of waiting for tests, which will add up significantly over the course of a year.

Note that the LazyDict created for this purpose is not thread safe. This is not a problem. The worst that could happen is that two threads simultaneously load the manifest. This would result in some wasted work. However, loading a manifest is idempotent, so there would be no other harmful effects.

### DIFF
--- a/dmcontent/utils.py
+++ b/dmcontent/utils.py
@@ -175,3 +175,32 @@ def count_unanswered_questions(service_attributes: "ContentManifest") -> typing.
                 unanswered_optional += 1
 
     return unanswered_required, unanswered_optional
+
+
+class LazyDict(collections.MutableMapping):
+    """
+    A dictionary for values that will be lazily evaluated the first time they are requested.
+    If a value is callable, then it will be called the first time that value is requested and the result cached.
+
+    This is probably not thread safe, so any callable inserted should be idempotent
+    """
+    def __init__(self, *args, **kw):
+        self._raw_dict = dict(*args, **kw)
+
+    def __getitem__(self, key):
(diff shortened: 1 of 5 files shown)


dmcontent/utils.py
```diff
@@ -175,3 +175,32 @@ def count_unanswered_questions(service_attributes: "ContentManifest") -> typing.
...
+        return iter(self._raw_dict)
+
+    def __len__(self):
+        return len(self._raw_dict)
+
+    def __setitem__(self, key, value):
+        self._raw_dict.__setitem__(key, value)
+
+    def __delitem__(self, value):
+        self._raw_dict.__delitem__(value)
```
lfdebrux, line 206: Doesn't matter too much but I think the argument should be `key` for clarity

```suggestion
    def __delitem__(self, key):
        self._raw_dict.__delitem__(key)
```

lfdebrux (approved):
Awesome stuff

dmcontent/utils.py
bjgill: Good catch, thanks. I think I misread the documentation. Fixed.

PR #140, opened 2021-03-17
Add support for file upload questions

alex9smith:
We want to use the `govukFileUpload` component[[1]] for our questions with `type: upload`[[2]].
Add a branch to `render_question` that generates parameters suitable for this component.

https://trello.com/c/7Hw9pH6X/855-2-add-type-upload-to-content-loader-renderquestion

[1]: https://design-system.service.gov.uk/components/file-upload/
[2]: https://github.com/alphagov/digitalmarketplace-frameworks/blob/a691598f98aa4a21c2615bdc07437898abe1591e/frameworks/digital-outcomes-and-specialists-5/questions/declaration/modernSlaveryStatement.yml

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '7.31.0'
+__version__ = '7.32.0'
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -135,6 +135,11 @@ def from_question(
             "macro_name": "govukCharacterCount",
             "params": govuk_character_count(question, data, errors, **kwargs)
         }
+    elif question.type == "upload":
+        return {
+            "macro_name": "govukFileUpload",
+            "params": govuk_file_upload(question, data, errors, **kwargs)
+        }
     elif question.type == "multiquestion":
         return dm_multiquestion(question, data, errors, **kwargs)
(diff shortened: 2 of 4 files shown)


lfdebrux:
This looks like a good start, but I think it needs a bit of extra work .

In the case where there is question advice that is usually HTML and quite long (it could contain links or a list even), so we should put that in the hint HTML slot I think (at least for now).

Also, the old frontend toolkit file upload macro adds an extra paragraph showing the user the URL of any file previously uploaded[[1], [2]], which the new component doesn't; can we hack that content in to the hint as well? We probably don't need to worry about the icon at least.

[1]: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/forms/macros/forms.html#L214
[2]: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/forms/upload.html#L57-L72

 I've attached a screenshot of an example of a complex file upload component from the DOS5 framework application (asking for a modern slavery statement document), which shows long question advice with links and lists, and the previously uploaded document.
 
 
Screenshot 2021-03-18 at 11 46 09

alex9smith:
Thanks Laurence, that's really helpful

alex9smith:
Questions with in-line HTML (eg. G-Cloud 12) and generated from YAML (eg. DOS5) now both render as:
image

(this is a screenshot from a hacked together route for testing so please excuse the dodgy breadcrumbs and the fact that this specific question should be shown as part of a multiquestion)

lfdebrux (approved):
Works great! Nice one 👍

PR #143, opened 2021-04-01
Allow for multiple conditional questions

domoscargin:
The question [modernSlaveryTurnover](https://github.com/alphagov/digitalmarketplace-frameworks/blob/main/frameworks/digital-outcomes-and-specialists-5/questions/declaration/modernSlaveryTurnover.yml) contains multiple, un-nested conditional questions.

We've (probably sensibly) allowed for only one conditional question.

This PR allows the `dm_multiquestion` function to create a list of conditional questions, and handles this case in the `render` function.

I've flagged this question as one which may benefit from some UR: https://trello.com/c/JPQmbdfD/9-better-way-of-handling-the-modern-slavery-question

### DIFF
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -378,9 +378,12 @@ def dm_multiquestion(
 
             for item in items:
                 if item["value"] in followups:
-                    followup_q = question.get_question(followups[item["value"]][0])
+                    followup_items = []
+                    for followup_id in followups[item["value"]]:
+                        followup_q = question.get_question(followup_id)
+                        followup_items.append(from_question(followup_q, data, errors, is_page_heading=False))
                     item["conditional"] = {
-                        "html": from_question(followup_q, data, errors, is_page_heading=False)
+                        "html": followup_items
                     }
 
     return to_render
@@ -589,11 +592,7 @@ def render(ctx, obj, *, question=None) -> Markup:
                 def visit(inner_obj):
                     if isinstance(inner_obj, dict):
                         for k, v in inner_obj.items():
-                            if (
-                                k == "html"
-                                and isinstance(v, dict)
-                                and "macro_name" in v
-                            ):
+                            if k == "html":
                                 inner_obj["html"] = render(ctx, v)
                             else:
                                 visit(v)
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -378,10 +378,10 @@ def dm_multiquestion(
 
             for item in items:
                 if item["value"] in followups:
-                    followup_items = []
-                    for followup_id in followups[item["value"]]:
-                        followup_q = question.get_question(followup_id)
-                        followup_items.append(from_question(followup_q, data, errors, is_page_heading=False))
+                    followup_items = [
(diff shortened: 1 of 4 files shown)


dmcontent/govuk_frontend.py
```diff
@@ -377,10 +377,20 @@ def dm_multiquestion(
...
-                    followup_q = question.get_question(followups[item["value"]][0])
-                    item["conditional"] = {
-                        "html": from_question(followup_q, data, errors, is_page_heading=False)
-                    }
+                    if len(followups[item["value"]]) > 1:
+                        number_of_followups = len(followups[item["value"]])
+                        followup_items = []
+                        for i in range(number_of_followups):
+                            followup_q = question.get_question(followups[item["value"]][i])
+                            followup_items.append(from_question(followup_q, data, errors, is_page_heading=False))
```
bjgill, line 385: Could we use a generator to make the logic here clearer?

dmcontent/govuk_frontend.py
```diff
@@ -377,10 +377,20 @@ def dm_multiquestion(
...
+                        item["conditional"] = {
+                            "html": followup_items
+                        }
+                    else:
```
bjgill, line 389: Why is `number_of_followups == 1` a special case?

bjgill (commented):
Looks plausible. There was something about the code I didn't understand - two questions below.

dmcontent/govuk_frontend.py
lfdebrux, line 385: In Python `for` loops can iterate over the elements in a list without needing to use `range`

```suggestion
                        followup_items = []
                        for followup_id in followups[item["value"]]:
                            followup_q = question.get_question(followup_id)
                            followup_items.append(from_question(followup_q, data, errors, is_page_heading=False))
```

dmcontent/govuk_frontend.py
domoscargin: Whelp, I got some failed tests when I didn't account for this, but in retrospect, I should've just adjusted the tests...

dmcontent/govuk_frontend.py
domoscargin: I've cleaned this up a bit - thanks @lfdebrux

dmcontent/govuk_frontend.py
```diff
@@ -378,9 +378,12 @@ def dm_multiquestion(
 
             for item in items:
                 if item["value"] in followups:
-                    followup_q = question.get_question(followups[item["value"]][0])
+                    followup_items = []
+                    for followup_id in followups[item["value"]]:
+                        followup_q = question.get_question(followup_id)
+                        followup_items.append(from_question(followup_q, data, errors, is_page_heading=False))
```
bjgill, line 384:
```suggestion
                    followup_items = [
                        from_question(question.get_question(followup_id), data, errors, is_page_heading=False)
                        for followup_id in followups[item["value"]]
                    ]
```

Your new version is much clearer. Could we use a generator to make it clearer still?

lfdebrux:
As Brett is on leave for a while I'll get this PR merged

bjgill (approved):
Looks good.

I don't know what effect this will have on our types/type hints. But they're already very broken, so this shouldn't make it any worse. :shipit:

PR #146, opened 2021-04-08
Fix type hints in dmcontent/govuk_frontend.py

bjgill:
Trello: https://trello.com/c/DZ28pqhm/90-try-to-add-type-hints-to-content-loader

Content loader is quite complex. We'd like to use mypy to check that we're not doing anything wrong. This PR is a pre-requisite that fixes the existing mypy failures in `dmcontent/govuk_frontend.py`.

We've not changed the type signature or behaviour of any methods used by the frontend apps. So this should have no effect outside this repo.

See individual commits for details of what (and why) this PR does.

### DIFF
--- a/dmcontent/govuk_frontend.py
+++ b/dmcontent/govuk_frontend.py
@@ -46,7 +46,7 @@ Read the docstring for `from_question()` for more detail on how Questions are
 handled.
 """
 
-from typing import List, Optional, Set, TYPE_CHECKING
+from typing import List, Optional, Set, Union, TYPE_CHECKING
 
 import jinja2
 from jinja2 import Markup, escape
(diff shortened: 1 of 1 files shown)


dmcontent/govuk_frontend.py
```diff
@@ -64,10 +64,12 @@
 # set this in your app to change the behaviour of this code.
 govuk_frontend_version = (2, 13, 0)
 
+MultiQuestionToRender = List[Union[dict, Markup]]
```
lfdebrux, line 67: Rather than making a type that is specific to multiquestions, I would make a type `Renderable = Union[dict, str, Markup, List[Union[dict, str, Markup]]`, as that describes everything that can be produced by `from_question` or handled by `render`

alex9smith: Good idea - added in f24d2ae!

lfdebrux (approved):
Looks good, 👍  for doing this!

PR #147, opened 2021-04-08
Fix type hints in dmcontent/questions.py

alex9smith:
Trello: https://trello.com/c/DZ28pqhm/90-try-to-add-type-hints-to-content-loader

Content loader is quite complex. We'd like to use mypy to check that we're not doing anything wrong. This PR is a pre-requisite that fixes the existing mypy failures in `dmcontent/questions.py`.

This is slightly more complex than #146 in that we've used `TypeVars` to show that methods on the base classes can be called by subclasses and return the type of the sub class. This is similar to the way typing is done for generics - the [mypy docs](https://mypy.readthedocs.io/en/stable/generics.html) have a pretty good explanation.

See individual commits for details of what (and why) this PR does.

### DIFF
--- a/dmcontent/questions.py
+++ b/dmcontent/questions.py
@@ -12,7 +12,8 @@ from .formats import format_price
 from .govuk_frontend import get_href
 from .utils import TemplateField, drop_followups, get_option_value
 
-T = TypeVar("T", bound="Question")
+TQuestion = TypeVar("TQuestion", bound="Question")
+TMultiquestion = TypeVar("TMultiquestion", bound="Multiquestion")
 
 
 class Question(object):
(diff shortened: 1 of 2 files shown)


dmcontent/questions.py
```diff
@@ -26,15 +28,20 @@ def __init__(self, data, number=None, _context=None):
...
+    def filter(self: T, context, dynamic=True, inplace_allowed: bool = False) -> Optional[T]:
         if not self._should_be_shown(context):
             return None
 
         if inplace_allowed:
             self._context = context
             return self
         else:
-            return ContentQuestion(self._data, number=self.number, _context=context)
+            content_question = ContentQuestion(self._data, number=self.number, _context=context)
```
lfdebrux, line 40: Instead of using the factory here could we just use `self.__class__` as a constructor?

lfdebrux (approved):
One small suggestion, but these changes makes sense to me 👍

dmcontent/questions.py
bjgill: Good point. And we also found another place where `ContentQuestion` was unnecessary

PR #159, opened 2021-06-28
Fix bug when generating list HTML

alex9smith:
Fix a bug where invalid HTML was generated when a HTML list was put in one of a question's Markdown fields (eg. `essentialRequirementsMet`[1])

This was because the markdown preprocessor (`markdown.preprocessors.HtmlBlockPreprocessor`) replaces raw HTML content with a placeholder value wrapped in an empty `<p></p>`. There is a postprocessor (`markdown.postprocessors.RawHtmlPostprocessor`) that finds the placeholders and removes the empty `<p></p>`, however we have a custom [`Treeprocessor`](https://github.com/alphagov/digitalmarketplace-content-loader/blob/main/dmcontent/markdown.py#L8) which added on the `govuk-body` class to these empty `<p>` tags and meant that they weren't removed by the post processor.

This meant that we had list elements wrapped in `<p>` tags in the output HTML which isn't valid HTML.

Add a check to `dmcontent.markdown._ClassAddingTreeprocessor` that ensures it doesn't add classes to these placeholder elements.

[1]:https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists-5/questions/brief-responses/essentialRequirementsMet.yml

https://trello.com/c/OMcXpf8k/42-2-content-loader-can-generate-invalid-list-html

### DIFF
--- a/dmcontent/__init__.py
+++ b/dmcontent/__init__.py
@@ -2,4 +2,4 @@ from .content_loader import ContentLoader
 from .errors import ContentTemplateError, QuestionNotFoundError
 from .questions import ContentQuestion
 
-__version__ = '8.2.0'
+__version__ = '8.2.1'
--- a/dmcontent/markdown.py
+++ b/dmcontent/markdown.py
@@ -21,8 +21,15 @@ class _ClassAddingTreeprocessor(Treeprocessor):
 
         # mutate each affected element in-place
         if element.tag in self._mapping:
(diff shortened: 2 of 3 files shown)


gidsg (approved):
Nice one 🏷️ 👍

alex9smith:
Ah, this will also need a version bump

PR #161, opened 2021-06-30
Fix bug when number field value is 0

domoscargin:
In some cases, we may pass 0 in a govukInput field and expect it to be a valid answer.

This doesn't display as Jinja evaluates 0 as falsy in the template: https://github.com/alphagov/govuk-frontend/blob/b2687fe47f3eb37d8790f422edb520fc24b4e0cd/src/govuk/components/input/template.njk#L51

We can fix this by stringifying `int`s for govukInputs, since we're using an input of `type='text'` anyway.


alex9smith (approved):
👍

PR #162, opened 2021-07-17
Add integer check to Question._should_be_shown()

domoscargin:
Some questions are dependent on others having a specific answer, but are not direct conditionals (currently, this only
applies to `socialValueCriteriaOutcomes`, which requires that `socialWeighting` be between 10 and 20).

Previously, we were only checking that a value was equal to any in a list of strings (and using this exclusively for checking `lot`). This commit adds the ability to check a number range.


tests/test_questions.py
```diff
@@ -61,18 +61,44 @@ def test_question_filter_without_dependencies(self, filter_inplace_allowed):
...
+            {"on": "lot", "being": ["lot-1"]},
+            {"on": "number", "being": [{"min_value": 10, "max_value": 10, "type": "integer"}]},
+            {
+                "on": "other_number",
+                "being": [
+                    {"min_value": 10, "max_value": 10, "type": "integer"},
+                    {"min_value": 0, "max_value": 0, "type": "integer"}
+                ]
+            }
+        ])
```
bjgill, line 97: Does this test hit the new code that deals with number dependencies?

dmcontent/questions.py
```diff
@@ -41,10 +41,26 @@ def filter(self: TQuestion, context, dynamic=True, inplace_allowed: bool = False
...
         )
 
+    def _is_dependency_valid(self, key, condition, context):
+        is_valid = False
+        if key in context:
+            if context[key] in condition:
+                is_valid = True
+            else:
+                boolean_array = []
+                for cond in condition:
```
bjgill, line 55: If there are potentially multiple conditions, why are we using the singular `condition`?

dmcontent/questions.py
```diff
@@ -41,10 +41,26 @@ def filter(self: TQuestion, context, dynamic=True, inplace_allowed: bool = False
...
+                    if "type" in cond and cond["type"] == "integer":
+                        min_value = cond.get("min_value", 0)
+                        max_value = cond.get("max_value", min_value) + 1
+                        boolean_array.append(context[key] in range(min_value, max_value))
```
bjgill, line 59: The `+ 1` surprised me - I assume this is for making an inclusive check? Could we use the comparison operators to make this clearer - something like `min <= context[key] <= max`, perhaps?

dmcontent/questions.py
```diff
@@ -41,10 +41,26 @@ def filter(self: TQuestion, context, dynamic=True, inplace_allowed: bool = False
 
     def _should_be_shown(self, context):
         return all(
-            depends["on"] in context and (context[depends["on"]] in depends["being"])
+            self._is_dependency_valid(depends["on"], depends["being"], context)
             for depends in self.get("depends", [])
         )
 
+    def _is_dependency_valid(self, key, condition, context):
```
bjgill, line 48: I know that content loader's performance strongly affects the apps (see all the work done to enable in-place filtering). Can we improve our performance of handling something like:

```
                "on": "other_number",
                "being": [
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                    {"min_value": 0, "max_value": 0, "type": "integer"}
                ]
```

dmcontent/questions.py
```diff
@@ -41,10 +41,26 @@ def filter(self: TQuestion, context, dynamic=True, inplace_allowed: bool = False
...
+        is_valid = False
+        if key in context:
+            if context[key] in condition:
+                is_valid = True
+            else:
+                boolean_array = []
+                for cond in condition:
+                    if "type" in cond and cond["type"] == "integer":
```
bjgill, line 56: I have a nasty, suspicious mind. How would this cope with:

```
{"on": "lot", "being": ["typewriters", "keyboards"]},
```

bjgill (commented):
The approach is sensible - I agree that this is the right place to add this logic.

Given the importance of content loader to the apps, I have a couple of questions about performance and ability to cope with unexpected inputs.

dmcontent/questions.py
bjgill: (looks like some type hints might help mypy catch this issue)

domoscargin:
Hmm, this clearly needs more work. Rushed it. Going to convert to draft.

